PR #31, opened 2018-11-23
[BLOCK] (found no way yet to proxy) update start script to only use username password if supplied

Derryrover:
Edit: right now the problem i wrote this branch for doesnot occur anymore.
Also this fix breaks authentication.
I will continue on this branch if the problem comes back.

When proxying to local socker providing the usernmae password with every api call no results it very slow behaviour. It seems that when supplying username password to the api it will re-authenticate every time. This makes each api call very slow and difficult to program with. Anyway, when proxying to prod it seems sometimes needed to provide username password. I do not fully understand this behaviour yet, but in the meanwhile need this fix in order to develop faster


scripts/start.js
```diff
@@ -1,35 +1,35 @@
-'use strict';
```
Derryrover: It seems that the linter changed all single quotes to double.
The only real change is the if statment on line 64

PR #118, opened 2019-03-29
Make asset select box readOnly if no raster is selected.

GeOdin:
Make select asset readonly if no raster is selected
Gray out select asset if no raster is selected


GeOdin:
Also make sure that the drop down menu of the active component is always visible.

PR #121, opened 2019-04-16
Remco webpack4 typescript

remcogerlich:
Extracting and working with translations will be changed in the refactoring, current ones still work


package.json
```diff
@@ -1,235 +1,149 @@
...
+      "web.ts",
+      "ts",
+      "web.tsx",
+      "tsx",
       "json",
       "web.jsx",
       "jsx",
       "node"
+    ],
+    "watchPlugins": [
```
Derryrover: bij deze staat je username remco.gerlich nog in het padnaam.

PR #131, opened 2019-05-14
Go to different url when other organisation is selected.

GeOdin:


Derryrover:
Already in templates lizard-nxt

PR #132, opened 2019-05-20
[Done] Fix crashing submit raster modal by using animationData.default and give previewColor of ColorMapComponent an unique key

GeOdin:


remcogerlich:
👍

PR #134, opened 2019-06-03
Remco new form

remcogerlich:
Work in progress. First version works -- the functionality is there, making forms is _way_ easier now, but so far only TextInput is done. And no translations yet.

The code is in src/forms/ , an example form with some dummy fields is in src/data_management/rasters/NewRaster2.js.


remcogerlich:
Leftovers:
- There are still bugs if the last field of a wizardStyle form is disabled
- Want a way to initialize all fields with an object
- Want a way to make all fields readonly
- Text Inputs don't support multiline yet
- The big one: the existing raster forms must be replaced with the new one. Check that was have all features

Choosing to skip DateInput as I kept zoning out while trying to make it. It's not complicated but not required for now.

PR #135, opened 2019-06-12
add functionality to flush raster data

Derryrover:
Refactor duplicated code

### DIFF
--- a/src/data_management/rasters/Raster.js
+++ b/src/data_management/rasters/Raster.js
@@ -185,117 +185,75 @@ class Raster extends Component {
   }
 
   handleDeleteRasterClick() {
-    var toBeDeletedRasterNamesArray = [];
-    var toBeDeletedRasterUuidsArray = [];
-
-    this.state.checkboxes.forEach(function(checkbox) {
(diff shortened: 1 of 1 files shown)


src/data_management/rasters/Raster.js
```diff
@@ -184,71 +185,77 @@ class Raster extends Component {
...
+    });
+    Promise.all(fetches).then(values => {
+      // Refresh the page, so that the removed rasters are no longer visible
+      this.getRastersFromApi(
+        this.state.page,
+        this.state.searchTerms,
+        this.state.include3diScenarios
+      );
+    });
+    // TODO show user results of the fetch
```
hoanphungt: I see a lot of comments here. I suppose these comments are to implement a new feature of the data management and maybe you want to keep them here to do it later on?

Derryrover: Yep I think it would be nice to give the user some specific feedback about which rasters are deleted. 
It was however a bit too complicated at first.
But maybe we could have a simpler message just saying: "rasters deleted" or something

PR #136, opened 2019-06-13
Tom add slushbucket

Derryrover:


Derryrover:
should merge to branch of remco instead

PR #156, opened 2019-07-17
Added more explanation when create a new timeseries alarm

hoanphungt:


src/alarms/notifications/NewNotification.js
```diff
@@ -880,7 +880,7 @@ class NewNotification extends Component {
                                   "notifications_app.select_timeserie" +
                                   "_input"
                                 }
-                                placeholder={"Click to select timeseries"}
+                                placeholder={"Select the timeseries in which the alarm will be added to"}
```
Derryrover: I would remove "in":
"Select the timeseries which the alarm will be added to"

hoanphungt:
The translation will be quick. Probably 15'. I am thinking creating a separate ticket to translate all placeholders in the management screen. Do you think it is necessary to do so? Also, I just check the management screen again. Seems like only here there are placeholders left, so probably not worth it to create a separate ticket for it

Derryrover:
@hoanphungt 
Hi Hoan,
Doesn't really mather to me if you create a seperate ticket or not.
Most important is that we do not forget about that this still has to be done.
So you can do it as part of the current ticket (and keep the current ticket open untill it is done)
Or create a new ticket an close the current ticket.

kr Tom

hoanphungt:
@Derryrover 
Hi Tom, 
It is already been translated so I closed this pull request and also the ticket for this. 
Kr, Hoan

PR #157, opened 2019-07-22
Tom alarm phonenumber

Derryrover:


src/alarms/contacts/NewContact.js
```diff
@@ -31,55 +31,82 @@ class NewContact extends Component {
...
+    const firstNameTranslation = this.props.intl.formatMessage({ id: "contacts_app.first_name_3" });
+    const lastNameTranslation = this.props.intl.formatMessage({ id: "contacts_app.last_name_3" });
+    const emailTranslation = this.props.intl.formatMessage({ id: "contacts_app.email_address_3" });
+    // const phoneNumberTranslation = this.props.intl.formatMessage({ id: "contacts_app.phone_number_3" });
+    const fieldsDidnotValidateTranslation =  this.props.intl.formatMessage({ id: "contacts_app.following_fields_didnot_validate" }) + ' \n '; 
+
+    let errorMessage = fieldsDidnotValidateTranslation;
+    let showErrorMessage = false;
+
+    if ( firstName.length == 0 ) {
```
hoanphungt: Should we use === instead of ==? and when do you use == and ===? For me, I would always use === except for some cases where only the == sign would give me the required result.

src/alarms/contacts/NewContact.js
```diff
@@ -31,55 +31,82 @@ class NewContact extends Component {
...
+    })
+    .then(data => {
+      history.push("/alarms/contacts");
+    });
   }
   render() {
     return (
       <div className={gridStyles.Container}>
-        <div className={`${gridStyles.Row}`}>
+        <div
```
hoanphungt: Why do you need this field here with "display:none"? I don't see anywhere in the code you want to display this field at later stages so it is quite redundant in my opinion. Maybe you have another intention for it :)

src/alarms/contacts/NewContact.js
Derryrover: True. Will change this. It is better to be stricter.

src/alarms/contacts/NewContact.js
Derryrover: Also true. It will never be visible for the user. It is for the translation engine so that it is able to extract the string. I will add a comment in the code.

PR #164, opened 2019-08-12
Translation for scenario management screen

hoanphungt:


src/data_management/scenarios/Scenarios.js
```diff
@@ -1,6 +1,7 @@
 import React, { Component } from 'react';
 import MDSpinner from "react-md-spinner";
 import { Scrollbars } from "react-custom-scrollbars";
+import { FormattedMessage, injectIntl } from "react-intl";
```
GeOdin: is 'injectIntl' also needed?

PR #165, opened 2019-08-12
improve wms layer screens

Derryrover:


src/data_management/wms_layers/WmsLayerForm.js
```diff
@@ -102,6 +102,16 @@ class WmsLayerFormModel extends Component {
     }
   }
 
+  optionsValidator = (value) => {
```
hoanphungt: I think the optionsValidator should also accept null value or empty string as this field is not mandatory to fill in for the user. What do you think? For now, if I don't want to configure this, I need to provide an empty json object {} or click on the X icon on the right to move on.

src/components/SearchBox.js
```diff
@@ -63,6 +63,7 @@ class SearchBox extends Component {
                 : { right: "6px" }
             }
             onClick={() => {
+              console.log('onclear searchbox clicked');
```
hoanphungt: console.log() should be removed if not necessary

src/data_management/wms_layers/ErrorOverlay.js
```diff
@@ -83,7 +83,7 @@ class ErrorOverlay extends Component {
     ) {
       return (
         <FormattedMessage
-          id="raster.post_metadata_success"
+          id="wms_layer_form.post_success"
           defaultMessage="Succes! Your raster meta-data was uploaded succesfully. You can add your data now, or do it later"
```
hoanphungt: you should change the default message here as well :)

PR #166, opened 2019-08-13
Add functionalities to scenarios management screen

hoanphungt:
Including:
- Pagination
- Searching
- Ordering
- Deleting

### DIFF
--- a/src/components/SearchBox.js
+++ b/src/components/SearchBox.js
@@ -6,15 +6,16 @@ import formStyles from "../styles/Forms.css";
 import clearInputStyles from "./ClearInputButton.css";
 
 class SearchBox extends Component {
+
   handleEnter(event) {
     if (event.keyCode === 13) {
       // 13 is keycode 'enter' (works only when current input validates)
-      this.props.handleSearch(this.props.searchTerms);
+      this.props.handleSearchEnter();
     }
   }
 
   render() {
-    const { searchTerm, handleSearch, intl } = this.props;
+    const { searchTerms, handleSearchEnter, handleSearchOnBlur, handleSearchClear, intl } = this.props;
 
     return (
       <div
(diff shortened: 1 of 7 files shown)


src/components/SearchBox.js
```diff
@@ -53,6 +54,9 @@ class SearchBox extends Component {
             }}
             onKeyUp={e => this.handleEnter(e)}
             title={intl.formatMessage({ id: "Search in name and description" })}
+            onBlur={e => {
```
Derryrover: With the changes in this file searchbox.js you will encounter two problems:
- It will break existing search functionality on rasters
- It will get a merge conflict with master
Maybe it would be better to revert this file to original.
Then merge master in "hoan-scenario_basic-setup"
Then merge "hoan-scenario_basic-setup" in this branch.
Then see if the searchbar.js component fullfills your needs.
If not you can change it, but also keep testing raster page.

src/data_management/scenarios/PaginationBar.js
```diff
@@ -0,0 +1,81 @@
+import React, { Component } from "react";
+import styles from "./PaginationBar.css";
+
+class PaginationBar extends Component {
+    state = {
+        inputPage: ""
```
Derryrover: Probably not needed anymore.
Do you know where it is used?

src/components/SearchBox.js
hoanphungt: Ah these are not changes made by me. I had the previous master branch and copied the searchbar.js from madeleine's branch to use it here. I probably would only need to delete this file and merge master back into this branch.

src/data_management/scenarios/PaginationBar.js
hoanphungt: Ah used it in componentDidMount and componentWillReceiveProps but I do think that it is not needed. I will just remove them all and test it if it is still needed.

src/components/SearchBox.js
hoanphungt: Ah you are right. Your work flow is more correct. I should merge master into the base branch first and then merge the base branch into this branch.

src/data_management/scenarios/PaginationBar.js
hoanphungt: Test it and it is not needed anymore so I removed it.

PR #173, opened 2019-09-18
WMS improvement

hoanphungt:
- Update wms_url field


src/data_management/wms_layers/WmsLayerForm.js
```diff
@@ -179,11 +177,11 @@ class WmsLayerFormModel extends Component {
     const placeholderWmsLayerName = intl.formatMessage({ id: "placeholder_wms_layer_name" });
     const placeholderDescription = intl.formatMessage({ id: "placeholder_description" });
     const placeholderSlug = intl.formatMessage({ id: "placeholder_slug" });
-    const placeholderGetFeatureInfo = intl.formatMessage({ id: "placeholder_get_feature_info" });
-    const placeholderMinZoom = intl.formatMessage({ id: "placeholder_min_zoom" });
+    // const placeholderGetFeatureInfo = intl.formatMessage({ id: "placeholder_get_feature_info" });
```
GeOdin: Letting the error messages pass silently by commenting out these constants may not be the best way.
We use these arguments in the WMS Form page (#139-152).

It would be better to add them to the translations.

src/data_management/wms_layers/WmsLayerForm.js
```diff
@@ -114,7 +112,7 @@ class WmsLayerFormModel extends Component {
           organisation: validatedData.selectedOrganisation,
           name: validatedData.wmsLayerName,
           description: validatedData.description,
-          url: validatedData.wmsLayerUrl,
+          wms_url: validatedData.wmsLayerUrl,
```
GeOdin: Good one! :+1:

src/data_management/wms_layers/WmsLayerForm.js
hoanphungt: These constants are currently not used in the WMS Form Page. The GetFeatureInfo is a checkbox so there is no placeholder. The tiled is commented out in the code so placeholderTiled should also be commented out. Only the placeholderMinZoom, we need to add translation for it as well just like the case of placeholderMaxZoom. I will make changes.

PR #185, opened 2019-10-14
Fix organisation uuid is null

hoanphungt:


src/data_management/wms_layers/WmsLayer.js
```diff
@@ -76,7 +76,7 @@ class WmsLayer extends Component {
             .includes(searchContains.toLowerCase()))
           &&
             // also sort on organisation, but remove slash from organisation uuid
-            (e.organisation.uuid.split('-').join('')) === selectedOrganisationUuid
+            (e.organisation && e.organisation.uuid.split('-').join('')) === selectedOrganisationUuid
```
GeOdin: Good idea! :+1:

PR #187, opened 2019-10-17
No longer filter on admin and supplier for showing organisations on h…

GeOdin:
…omepage of lizard-management-client.

### DIFF
--- a/src/reducers.js
+++ b/src/reducers.js
@@ -75,9 +75,10 @@ function organisations(
         available: action.data
           .filter(e => {
             return (
+              e.roles.find(e => e === "manager") ||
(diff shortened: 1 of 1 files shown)


src/reducers.js
```diff
@@ -76,7 +76,8 @@ function organisations(
           .filter(e => {
             return (
               e.roles.find(e => e === "admin") ||
-              e.roles.find(e => e === "supplier")
+              e.roles.find(e => e === "supplier") ||
+              e.roles.find(e => e === "user")
```
hoanphungt: I think we also need to add "manager" role into the filter as well. The rest is good. I also tested it.

PR #190, opened 2019-10-23
[DO-NOT-MERGE] Create new form for alarm management page

hoanphungt:


src/alarms/App.js
```diff
@@ -8,10 +8,11 @@ import { App as NewContactApp } from "./contacts/NewContact";
 import { App as NewNotificationApp } from "./notifications/NewNotification";
 import { App as NewTemplateApp } from "./alarmtemplates/NewTemplate";
 import { App as NotificationsApp } from "./notifications/App";
+import { App as EditNotificationApp } from "./notifications/EditNotification";
 import { Detail as AlarmGroupsDetail } from "./alarmgroups/Detail";
 import { Detail as AlarmTemplatesDetail } from "./alarmtemplates/Detail";
 import { Detail as ContactDetail } from "./contacts/Detail";
-import { Detail as NotificationsDetail } from "./notifications/Detail";
+// import { Detail as NotificationsDetail } from "./notifications/Detail";
```
remcogerlich: Delete code instead of commenting it out, that is what Git is for :-)

src/alarms/notifications/EditNotification.js
```diff
@@ -0,0 +1,69 @@
+import React, { Component } from "react";
+import { connect } from "react-redux";
+import { withRouter } from "react-router-dom";
+import MDSpinner from "react-md-spinner";
+
+import { NotificationForm } from "./NotificationForm";
+
+class EditNotificationModel extends Component {
```
remcogerlich: I'd like a comment saying what kind of props this component has (e.g. what format 'match' has). Or better, switch to Typescript and define it.

src/alarms/notifications/EditNotification.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    componentDidMount() {
+        const { match, alarmType } = this.props;
+
+        (async () => {
+            const url = (alarmType === "RASTERS") ? (
+                `/api/v4/rasteralarms/${match.params.id}/`
+            ) : (
+                `api/v4/timeseriesalarms/${match.params.id}/`
```
remcogerlich: One has a starting /, the other doesn't.

remcogerlich: I think this code should be in constructor, no need to have it in a separate componentDidMount function.

src/alarms/notifications/EditNotification.js
```diff
@@ -0,0 +1,69 @@
...
+            }).then(response => response.json());
+
+            this.setState({ currentNotification })
+        })();
+    }
+
+    render() {
+        if (
+            this.state.currentNotification &&
+            this.props.organisations.isFetching === false
```
remcogerlich: Why does this depend on organisations.fetching?

src/alarms/notifications/EditNotification.js
```diff
@@ -0,0 +1,69 @@
...
+const mapStateToProps = (state) => {
+    return {
+        organisations: state.organisations,
+        alarmType: state.alarmType
+    };
+};
+
+const App = withRouter(connect(mapStateToProps)(EditNotificationModel));
+
+export { App };
```
remcogerlich: I think it is confusing that all these screens are called "App" and then need to be renamed when importing them elsewhere, why not just export default so they can be imported as EditNotification?

src/alarms/notifications/ErrorOverlay.js
```diff
@@ -0,0 +1,216 @@
...
+    console.log('this.props.errorMessage', this.props.errorMessage, this.props.currentNotification);
+    if (
+      (this.props.errorMessage.status === 201 ||
+        this.props.errorMessage.status === 200) &&
+      this.props.currentNotification
+    ) {
+      return true;
+    } else {
+      return false;
+    }
```
remcogerlich: I don't like if statements that return true or false. Returning the value of the if () expression is equivalent, maybe cast to a bool with !!.

src/alarms/notifications/ErrorOverlay.js
```diff
@@ -0,0 +1,216 @@
...
+      animationData = animationSucces.default;
+    } else {
+      animationData = animationError.default;
+    }
+    return animationData;
+  }
+
+  // Different buttons based on succes or fail of a request
+  succesButtons() {
+    console.log('this.props.errorMessage', this.props.errorMessage, this.props.currentNotification);
```
remcogerlich: Console.log left in the code.

src/alarms/notifications/ErrorOverlay.js
```diff
@@ -0,0 +1,216 @@
...
+    ) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  // Generate message based on returned API status code
+  whichMessage() {
+    if (
```
remcogerlich: If you define 'const statusCode = this.props.errorMessage.status' here, then the checks below are shorter...

src/alarms/notifications/ErrorOverlay.js
```diff
@@ -0,0 +1,216 @@
...
+      this.props.errorMessage.status === 201 ||
+      this.props.errorMessage.status === 200
+    ) {
+      return (
+        <FormattedMessage
+          id="notification_form.post_success"
+          defaultMessage="Succes! Your alarm notification was uploaded succesfully."
+        />
+      );
+    } else if (this.props.errorMessage.status.toString().startsWith(4)) {
```
remcogerlich: And this could be "if (400 <= statusCode && statusCode < 500)" which I consider cleaner than casting to a string.

src/alarms/notifications/ErrorOverlay.js
```diff
@@ -0,0 +1,216 @@
...
+    ) {
+      animationData = animationSucces.default;
+    } else {
+      animationData = animationError.default;
+    }
+    return animationData;
+  }
+
+  // Different buttons based on succes or fail of a request
+  succesButtons() {
```
remcogerlich: I think "showSuccessButtons" is a better name for a functioning returning a boolean; with this name it sounds like it returns the buttons themselves.

src/alarms/notifications/GroupAndTemplateSelect.js
```diff
@@ -47,12 +47,24 @@ class GroupAndTemplateSelect extends Component {
...
+  getMessageName(messages, messageId) {
+    if (messages.length === 0) return null;
+    const selectedMessage = messages.find(message => message.id === messageId);
+    return selectedMessage.name;
+  }
+  getGroupName(groups, groupId) {
+    if (groups.length === 0) return null;
+    const selectedGroup = groups.find(group => group.id === groupId);
+    return selectedGroup.name;
+  }
```
remcogerlich: For both these: handle the case that the message isn't found, even if you know it will be (this is called "defensive programming").

src/alarms/notifications/NewNotification.js
```diff
@@ -1,120 +1,41 @@
...
-          name="snoozing"
-          title="Snoozing"
-          />
-          <TextInput
-          name="recipients"
-          title="Recipients"
-          />
-        </ManagementForm>
+    if (
+      this.props.organisations.isFetching === false
```
remcogerlich: Why does this depend on organisations being fetched?

src/alarms/notifications/NotificationForm.js
```diff
@@ -0,0 +1,430 @@
...
+import ErrorOverlay from "./ErrorOverlay.js";
+
+import ManagementForm from "../../forms/ManagementForm";
+import TextInput from "../../forms/TextInput";
+import { minLength, required } from "../../forms/validators";
+import SelectBox from "../../forms/SelectBox";
+import TimeseriesSelection, { timeseriesChosen } from "../../forms/TimeseriesSelection";
+import ThresholdsSelection from "../../forms/ThresholdsSelection";
+import Snoozing from "../../forms/Snoozing";
+import Recipients, { recipeintsValidator } from "../../forms/Recipients";
```
remcogerlich: Typo in "recipeintsValidator"

src/alarms/notifications/NotificationForm.js
```diff
@@ -0,0 +1,430 @@
...
+        body: JSON.stringify(body),
+      };
+
+      fetch(url, opts)
+        .then(responseParsed => {
+          this.handleResponse(responseParsed);
+          return responseParsed.json();
+        })
+        .then(parsedBody => {
+          console.log("parsedBody", parsedBody);
```
remcogerlich: console.log

src/alarms/notifications/NotificationForm.js
```diff
@@ -0,0 +1,430 @@
...
+                "Before"
+              ) || "After"
+            }
+          />
+          <RelativeField
+            name="relativeStart"
+            title="Relative start"
+            subtitle="Optional: Select the relative start of the simulation period"
+            validators={[durationValidator()]}
+            initial={
```
remcogerlich: I thought the Before/After select box was going to be part of the RelativeField. I think that is more logical than this situation. Then the calculation to number of seconds could be done in the field instead of here in onSubmit, and it looks more logical for the user I think.

src/components/SelectTimeseriesSearch.js
```diff
@@ -0,0 +1,186 @@
...
+    this.setState({ mustShowChoices: true, query: e.target.value });
+    this.props.updateModelValue(e.target.value);
+  }
+  // setQuery(props) {
+  //   if (this.props.choice) {
+  //     this.setState({
+  //       query: this.props.transformChoiceToDisplayValue(this.props.choice)
+  //     });
+  //   }
+  // }
```
remcogerlich: Here (and above) code in comments

src/alarms/notifications/EditNotification.js
hoanphungt: It doesn't depend on organisations. So I deleted it.

src/alarms/notifications/EditNotification.js
hoanphungt: Just to keep it consistent with other screens in the alarm management. I personally don't like to name them this way but rename all of them seems not relevant to the purpose of this ticket.

src/alarms/notifications/EditNotification.js
hoanphungt: Moved to inside constructor and added /

src/alarms/notifications/EditNotification.js
hoanphungt: Comment added

src/alarms/notifications/NewNotification.js
hoanphungt: It does not depend on organisations being fetched. So I removed this part.

src/alarms/notifications/GroupAndTemplateSelect.js
hoanphungt: Nice! Thanks for pointing it out. I now also handle the case if message isn't found.

src/alarms/notifications/NotificationForm.js
hoanphungt: Fixed typo

src/alarms/notifications/NotificationForm.js
hoanphungt: Removed!

src/components/SelectTimeseriesSearch.js
hoanphungt: Also removed!

src/alarms/notifications/NotificationForm.js
hoanphungt: I think its a more logical place to have them too and it's quite a big change. I managed to do it (combine before/after select box into relative field and move all the logic in onSubmit to here as well. The most confusing part of this step to me is that I need to convert the duration value (from the API call or from the input) from a string to duration object and to number of seconds back and forth quite many times. Please have a look at them in the new commits.

remcogerlich: The idea should be that the field keeps its own value in whatever format is easiest for the field to use; and it should be translated to and from the APIs format when the API response is received or a new one is submitted, so once in each direction.

PR #208, opened 2019-11-13
Update font to use Montserrat as the new font for the Lizard Manageme…

GeOdin:
…nt Client and make the header and footer in uppercase (except for the breadcrumbs).

### DIFF
--- a/src/App.css
+++ b/src/App.css
@@ -2,6 +2,7 @@
   display: flex;
   min-height: 100vh;
   flex-direction: column;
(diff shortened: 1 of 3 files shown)


src/App.css
```diff
@@ -101,7 +102,7 @@
   border-radius: 2px;
 }
 .OrganisationLinkContainer {
-  width: 160px;
+  width: 260px;
```
hoanphungt: Nice that you make it bigger :)

PR #210, opened 2019-11-13
Madeleine update tile

GeOdin:


src/components/AppIcon.css
```diff
@@ -23,13 +21,16 @@
...
+  width: 100%;
   padding-left: 10px;
   padding-right: 10px;
-  font-size: 18px;
-  width: 100%;
-  text-align: center;
   color: #333;
+  font-family: Montserrat;
+  font-size: 17px;
+  font-color: #FFFFFF;
```
hoanphungt: What do you want to change with font-color? if it is to change the color of the font then color is the one you need and it's already set at line 28.

src/components/AppIcon.js
```diff
@@ -6,12 +6,10 @@ import { withRouter } from "react-router-dom";
 
 class AppIcon extends Component {
   render() {
-    const { src, title, subTitle, handleClick } = this.props;
+    const { title, bgImage, handleClick } = this.props;
     return (
-      <div className={styles.AppIcon} onClick={handleClick}>
-        <img src={src} alt={title} className={styles.Img} />
+      <div className={styles.AppIcon} style={{backgroundImage: `url(${bgImage})`}} onClick={handleClick}>
```
hoanphungt: Nice that you use backgroundImage here for the whole tile.

src/home/App.js
```diff
@@ -52,7 +52,7 @@ class App extends Component {
             defaultMessage="Data Management"
           />
         ),
-        bgImage: templateIcon
+        bgImage: templateTileImage
```
hoanphungt: I find this name is quite misleading, it should be something like dataManagementTileImage. You also have a templatesTileImage in Alarms.js

src/components/AppIcon.css
GeOdin: Good idea! It is indeed no longer necessary.

src/home/App.js
GeOdin: Thank you! :) Forgot to rename it.

GeOdin: You made my day. :)

GeOdin:
Thank you!

For me de font is actually the same as for the design of Lirry.
Could you pull the madeleine-redesign-management-client branch again and update it?
(`git pull origin madeleine-redesign-management-client` when on the branch)

PR #212, opened 2019-11-15
remove deploy submodule

Derryrover:


Derryrover:
@remcogerlich 
Hi Remco,
Just to be sure, could you review if you agree ?
I am going tocreate a similar PR for all clients.
Would be good if at least one gets a review

kr Tom

PR #213, opened 2019-11-15
Madeleine update raster table page and alarm templates

GeOdin:


src/data_management/rasters/Raster.js
```diff
@@ -517,12 +513,12 @@ class Raster extends Component {
                   : `${buttonStyles.Button} ${buttonStyles.Inactive}`
               }
               onClick={this.handleFlushDataRasterClick}
-              style={{ maxHeight: "36px", width: "204px",marginRight: "10px" }}
+              style={{ maxHeight: "36px", width: "213px",marginRight: "10px" }}
```
hoanphungt: I think you can make it a bit bigger because for now, if I select 10 rasters then the text will go slightly off the button

src/data_management/rasters/Raster.js
```diff
@@ -636,13 +632,13 @@ class Raster extends Component {
         </div>
 
         <div>
+          {htmlRasterTableHeader}
```
hoanphungt: I think the header should also be kept inside the Scrollbars, otherwise, the scrollbar will not work for small screen-size (this is a horizontal scrollbar - there is another vertical scrollbar for the raster table)

src/styles/RasterWmsTable.css
```diff
@@ -15,6 +15,8 @@
 }
 
 .tableHeader {
+  display: flex;
+  align-items: center;
```
hoanphungt: nice that you kept everything vertically align here in the table header

src/data_management/rasters/Raster.js
GeOdin: Thank you. :)

src/data_management/rasters/Raster.js
GeOdin: This scrollbar actually also functions as a vertical scrollbar for the whole table (header and body of the table) if the header is also put in the Scrollbar.
I added it within the Scrollbar again and made the overflow-y hidden.

PR #214, opened 2019-11-18
Madeleine refactor breadcrumbs

GeOdin:
I made a new component for the breadcrumbs.
I also improved the way the styling is done.

### DIFF
--- a/src/components/Breadcrumbs.js
+++ b/src/components/Breadcrumbs.js
@@ -1,7 +1,5 @@
 import React, { Component } from "react";
 import { NavLink } from "react-router-dom";
-// import onClickOutside from "react-onclickoutside";
-// import styles from "./Breadcrumbs.css";
 import styles from "./Breadcrumbs.css";
 import gridStyles from "./../styles/Grid.css";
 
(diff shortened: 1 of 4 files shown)


src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,52 @@
...
+
+
+class Breadcrumbs extends Component {
+  constructor(props) {
+    super(props);
+    this.uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i;
+  }
+  computeBreadcrumb() {
+    const { pathname } = this.props.location;
+    const splitPathnames = pathname.slice().split("/");
```
Derryrover: // splitted on "/", but string starts with "/" so first item becomes empty string

src/components/Breadcrumbs.js
```diff
@@ -0,0 +1,52 @@
...
+    return pathname === "/"
+      ? null
+      : splitPathnames.map((sp, i) => {
+        const to = `/${splitPathnames.slice(1, i + 1).join("/")}`;
```
Derryrover: to could be called url or relative url.
// splice from 1 and not 0 to remove first empty string from array that was caused by splitting on "/"

PR #216, opened 2019-11-19
Madeleine add icons empty pages

GeOdin:


src/alarms/contacts/App.js
```diff
@@ -158,240 +159,253 @@ class App extends Component {
...
+        {total === 0 ? (
+          <div className={styles.NoResults}>
+            <img src={contactsIcon} alt="Contacts" />
+            <h5>
+              <FormattedMessage
+                id="alarmtemplates_app.no_contacts_configured"//to do
+                defaultMessage="No contacts configured..."
+              />
+            </h5>
+          </div>
```
hoanphungt: Nice

src/data_management/scenarios/Scenarios.css
```diff
@@ -149,6 +149,21 @@
...
+  flex-direction: column;
+  justify-content: center;
+  height: 300px;
+  align-items: center;
+}
+.NoResults h5 {
+  color: #C1C1C1;
+  font-weight: 300;
+  margin: 30px;
+}
```
hoanphungt: I can see that this piece of CSS styling is used multiple times in the whole App (i.e. the alarm notifications, raster, wms layers, contacts, templates, groups ...). It would be nice if we can reuse them across the whole app.

src/data_management/scenarios/Scenarios.js
```diff
@@ -245,31 +246,31 @@ class Scenarios extends Component {
 
         const scenarioTableHeader = () => {
             return (
-                <div className={scenartioStyle.tableHeader}>
-                    <div className={scenartioStyle.tableCheckbox}>
+                <div className={scenarioStyle.tableHeader}>
```
hoanphungt: Thanks for fixing my spelling mistake :)

src/data_management/scenarios/Scenarios.js
```diff
@@ -278,61 +279,75 @@ class Scenarios extends Component {
...
+            if (scenarios.length === 0 && isFetching === false) {
+                return (
+                    <div className={scenarioStyle.NoResults}>
+                        <img src={scenariosIcon} alt="3Di scenarios" />
+                            <h5>
+                                <FormattedMessage
+                                    id="scenario.no_scenarios_configured"
+                                    defaultMessage="No 3Di scenarios configured..."
+                                />
+                            </h5>
```
hoanphungt: Nice catch :)

src/data_management/wms_layers/WmsLayer.js
```diff
@@ -292,108 +292,99 @@ class WmsLayer extends Component {
...
+                justifyContent: "center",
+                alignItems: "center",
+                minHeight: "450px"
+              }}
+            >
+              <MDSpinner />
+            </div>
+          ) : (
+            <div
+              style={{
```
hoanphungt: we should remove this line if this is empty inside

src/data_management/wms_layers/WmsLayer.js
```diff
@@ -292,108 +292,99 @@ class WmsLayer extends Component {
...
-                    </NavLink>
-                  </div>
-                  <div className={`${wmsLayerTableStyles.tableDescription}`}>
-                    <NavLink
-                      to={`/data_management/wms_layers/${wmsLayer.uuid}`}
-                      style={{
-                        color: "#333"
-                      }}
+
+          {total === 0 && isFetching === false ? (
```
hoanphungt: I see that we have different ways of showing the no results across different management screens. Sometimes, we check for total and isFetching, sometimes we check for only total and in the scenarios, we check for the length. Also, this part is quite similar among all the management screens so maybe we can refactor it into a separate component and reuse it. But I think it would be for another ticket since it can take quite a lot of time as well.

src/data_management/wms_layers/WmsLayer.js
```diff
@@ -292,108 +292,99 @@ class WmsLayer extends Component {
...
+            <div
+              className={styles.NoResults}
+              style={{
+                display: "flex",
+                justifyContent: "center",
+                alignItems: "center",
```
hoanphungt: I think we don't need these styling here since it's already been defined in the styles.NoResults. Only the minHeight, we need to define.

GeOdin: Good idea!

src/data_management/wms_layers/WmsLayer.js
GeOdin: I agree, I will make a new ticket for this.

src/data_management/scenarios/Scenarios.css
GeOdin: I agree. I will put this in the refactoring ticket. :+1:

PR #217, opened 2019-12-02
WMS - Add spatial bounds and download url fields to the WMS layer form

hoanphungt:


src/data_management/wms_layers/WmsLayerForm.js
```diff
@@ -370,6 +370,20 @@ class WmsLayerFormModel extends Component {
...
+          title={<FormattedMessage id="wms_layer_form.spatial_bounds" />}
+          subtitle={<FormattedMessage id="wms_layer_form.add_spatial_bounds" />}
+          initial = {
+            (
+              currentWmsLayer &&
+              currentWmsLayer.spatial_bounds
+            ) || null
+          }
+          validators={[
+            (fieldValue) => spatialBoundsValidator(fieldValue)
```
remcogerlich: `(param) => someFunc(param)` is exactly the same as just `someFunc`; so this can be `validators={[spatialBoundsValidator]}`

src/utils/getBoundsFromGeoServer.js
```diff
@@ -0,0 +1,32 @@
+import X2JS from 'x2js';
+
+export const getBoundsFromWmsLayer = (wmsSlug, wmsUrl, value, valueChanged) => {
+    // Use /proxy/ to avoid the request is blocked by CORS policy
+    // Note: in production, it will probably be working without the prefix /proxy/
+    const proxyUrl = `/proxy/${wmsUrl}`;
```
remcogerlich: If that comment is true, then this needs to be changed (we should only commit to master what works in production). But I think that on production this is also needed.

src/utils/getBoundsFromGeoServer.js
```diff
@@ -0,0 +1,32 @@
...
+
+    fetch(`${proxyUrl}/?request=getCapabilities`)
+        .then(response => {
+            if (!response.ok) throw new Error('Oops! Cannot get data from the GeoServer');
```
remcogerlich: This error should be handled somehow (e.g. show an error next to the button).

src/forms/SpatialBoundsField.tsx
```diff
@@ -0,0 +1,199 @@
...
+            east,
+            south,
+            west
+        } = fieldValue;
+
+        if (
+            !north ||
+            !east ||
+            !south ||
+            !west
```
remcogerlich: This is wrong as 0 is a valid value, and "somestring" is not.

I think the check should be on `typeof north === "number"`.

src/forms/SpatialBoundsField.tsx
```diff
@@ -0,0 +1,199 @@
...
+            wmsLayerUrl,
+        } = otherValues;
+
+        let north, east, south, west;
+
+        if (value) {
+            north = value.north ? value.north : "";
+            east = value.east ? value.east : "";
+            south = value.south ? value.south : "";
+            west = value.west ? value.west : "";
```
remcogerlich: Same here -- 0 is valid.

src/forms/SpatialBoundsField.tsx
```diff
@@ -0,0 +1,199 @@
...
+                    <div
+                        className={
+                            styles.DurationInputFields +
+                            " " +
+                            styles.DurationInputFieldDays +
+                            " " +
+                            styles.TextAlignCenter
+                        }
+                    >
+                        <label>North (&deg;)</label>
```
remcogerlich: Missing translation

src/forms/SpatialBoundsField.tsx
```diff
@@ -0,0 +1,199 @@
...
+                        <input
+                            id="north"
+                            type="number"
+                            className={
+                                formStyles.FormControl +
+                                " " +
+                                styles.TextAlignCenter
+                            }
+                            value={north}
+                            onChange={(e) => this.updateSpatialBounds('north', e.target.value)}
```
remcogerlich: Cool 
👍

src/forms/SpatialBoundsField.tsx
```diff
@@ -0,0 +1,199 @@
...
+                        display: (wmsLayerSlug && wmsLayerUrl) ? 'block' : 'none'
+                    }}
+                    onClick={() => getBoundsFromWmsLayer(
+                        wmsLayerSlug,
+                        wmsLayerUrl,
+                        value,
+                        valueChanged
+                    )}
+                >
+                    Get from GeoServer
```
remcogerlich: Missing translation

src/data_management/wms_layers/WmsLayerForm.js
hoanphungt: Nice

src/utils/getBoundsFromGeoServer.js
hoanphungt: Yeah you are right, because the request is sent to another server outside of Lizard so it is probably needed on production as well. I will remove this comment.

src/utils/getBoundsFromGeoServer.js
hoanphungt: I have a catch at line 30 that give user an alert message.

src/forms/SpatialBoundsField.tsx
hoanphungt: I changed the check to Number.isNaN(north) ... because north can only be NaN or a number while typeof NaN also === "number"

src/utils/getBoundsFromGeoServer.js
remcogerlich: Ah right, I thought it said console.error instead of alert. But alert() looks very ugly and the error message is likely to be hard to understand for a user, I think something like a short message "failed to get extent from geoserver" in red next to the button would be better.

src/forms/SpatialBoundsField.tsx
hoanphungt: same here, I changed the check to `Number.isFinite(value.north) ? value.north : "";`

src/utils/getBoundsFromGeoServer.js
hoanphungt: Nice! I removed the alert() and added an error message next to the button :)

PR #219, opened 2019-12-04
Fix alarm thresholds not lined up correctly when threshold name is too long

hoanphungt:
* Also add sorting functionality for thresholds by their value


src/forms/ThresholdsSelection.tsx
```diff
@@ -75,6 +77,24 @@ class ThresholdsSelectionInput extends Component<MyProps & InjectedIntlProps, My
...
+        } else {
+            this.setState({
+                thresholdOrderDirection: 'asc'
+            });
+            newThresholds = thresholds.sort((a, b) => b.value - a.value);
+        };
+        this.props.valueChanged({
+            ...this.props.value,
+            thresholds: newThresholds
+        });
```
remcogerlich: I think this implements the wrong thing. What the ticket means is that the thresholds should *always be shown* in sorted order (ascending), not that there should be a button to let the user sort them.

PR #220, opened 2019-12-06
Fix alarm thresholds not lined up correctly and sort threshold values

hoanphungt:
- Fix alarm thresholds not lined up correctly
- Sort threshold values in ascending order

### DIFF
--- a/src/forms/ThresholdsSelection.css
+++ b/src/forms/ThresholdsSelection.css
@@ -1,10 +1,10 @@
-th {
+.ThresholdsTable th {
   height: 50px;
 }
 
-td {
+.ThresholdsTable td {
   max-width: 200px;
-  padding: 4px 10px;
+  padding: 4px 10px 4px 0;
 }
 
 .Comparision {
--- a/src/forms/ThresholdsSelection.css
+++ b/src/forms/ThresholdsSelection.css
@@ -1,8 +1,10 @@
-.Threshold {
-  width: 100%;
-  border: 1px solid #a5a5a5;
-  padding: 5px;
-  margin: 0 0 10px 0;
+th {
+  height: 50px;
+}
+
+td {
+  max-width: 200px;
+  padding: 4px 10px;
 }
 
 .Comparision {
@@ -11,45 +13,28 @@
   color: #707070;
 }
 
-.ThresholdInput {
-  display: flex;
-}
-
-.ThresholdInput label {
+.ThresholdLabel {
   color: #707070;
+  height: 50px;
   margin: 10px 0;
 }
 
-.ThresholdInput input {
-  margin: 6px 0;
-}
-
-.ThresholdValueInput {
-  max-width: 160px;
-  display: flex;
-  flex-direction: column;
-}
-
-.ThresholdNameInput {
-  max-width: 160px;
-  display: flex;
-  flex-direction: column;
-  margin-left: 10px;
-}
-
-.ThresholdValues {
+.ThresholdValue {
   font-weight: bold;
   color: #707070;
 }
 
-.ThresholdNames {
+.ThresholdName {
   font-weight: bold;
   color: #707070;
-}
-
-.ThresholdNames div {
   display: flex;
   justify-content: space-between;
+  align-items: center;
+}
+
+.ThresholdNameText {
+  width: 90%;
+  word-wrap: break-word;
 }
 
 .ThresholdDelete {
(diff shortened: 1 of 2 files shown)


src/forms/ThresholdsSelection.tsx
```diff
@@ -92,120 +91,134 @@ class ThresholdsSelectionInput extends Component<MyProps & InjectedIntlProps, My
...
+                            />
+                        </td>
+                        <td className={styles.ThresholdLabel}>
+                            <FormattedMessage
+                                id="notifications_app.threshold_name"
+                                defaultMessage="Name"
+                            />
+                        </td>
+                    </tr>
+                    {thresholds.sort((a, b) =>
```
Derryrover: I like it how you combined two seperate loops and intoo this single loop.
Defenitely an improvement !
The previous code was completely wrong here and you noticed it correctly.

src/forms/ThresholdsSelection.tsx
```diff
@@ -92,120 +91,134 @@ class ThresholdsSelectionInput extends Component<MyProps & InjectedIntlProps, My
...
-                <div className={styles.AddThreshold}>
-                    <label htmlFor="comparison" className={styles.Comparision}>
-                        <FormattedMessage
-                            id="notifications_app.comparison"
-                            defaultMessage="Comparison"
-                        />
-                    </label>
-                    <div className={styles.ThresholdInput}>
-                        <div className={styles.ThresholdValueInput}>
+            <table>
```
Derryrover: I notice you use a table here.
I am myself not a fan of the table, but this is a good use-case for it.
I am interested why did you choose table over div with flexbox?

src/forms/ThresholdsSelection.tsx
```diff
@@ -92,120 +91,134 @@ class ThresholdsSelectionInput extends Component<MyProps & InjectedIntlProps, My
...
-                                id="value"
-                                value={thresholdValue}
-                                className={formStyles.FormControlSmall}
-                                onChange={this.handleChangeThresholdValue}
-                            />
-                        </div>
-                        <div className={styles.ThresholdNameInput}>
+                        </td>
+                        <td>
                             <button
```
Derryrover: These buttons do either not below in the table or otherwise belong in the header of the table.
In my opinion they should not be in the body.

src/forms/ThresholdsSelection.tsx
```diff
@@ -92,120 +91,134 @@ class ThresholdsSelectionInput extends Component<MyProps & InjectedIntlProps, My
...
-                                            &#10007;
-                                      </span>
-                                    </div>
-                                ))}
-                            </div>
+                        </td>
+                    </tr>
+                    <tr>
+                        <td className={styles.ThresholdLabel}>
+                            <FormattedMessage
```
Derryrover: These belong in the header of the table

src/forms/ThresholdsSelection.css
```diff
@@ -1,8 +1,10 @@
-.Threshold {
-  width: 100%;
-  border: 1px solid #a5a5a5;
-  padding: 5px;
-  margin: 0 0 10px 0;
+th {
```
Derryrover: Styling th and td like this has an effect on all th and td of the whole page.
Somehow the react css import doesnot scope those.
I struggle to find the documentation on this unfortuanedly..
But you can find the prove of it by inspecting that the css selectors in the browser are not prefixed.
Very confusing part of css modules...

A solution would be to use class selectors in combination with element selectors.
Check css combinators.
Or you could use just only classes.
Please let me know if you have questions about this.

src/forms/ThresholdsSelection.tsx
hoanphungt: hmm I am not the one who comes up with the idea of using table here but Remco :) and I also want to try it with table and it works well I think. I think it's actually better than using div with flexbox here.

src/forms/ThresholdsSelection.tsx
hoanphungt: Moved to table header. To make them vertically align with other items in the column, I keep them in the table as well.

src/forms/ThresholdsSelection.css
hoanphungt: Ah you are right. This is so annoying indeed :(. I added a class selector for it.

src/forms/ThresholdsSelection.tsx
hoanphungt: Moved to header of table

PR #223, opened 2019-12-17
Align user and support buttons

hoanphungt:


GeOdin:
The width is no longer necessary.

PR #504, opened 2020-11-12
Tom raster list styling feedback1

Derryrover:


Derryrover:
Should be pr to feature branch

PR #508, opened 2020-11-12
Hoan improve form - part 1

hoanphungt:


Derryrover (approved):
Looks good!
I see you added a lot of types which is good.
Some of the components we now re-use we might refactor or write from scratch lateron.

PR #517, opened 2020-11-19
Upload raster data

hoanphungt:


hoanphungt:
@Derryrover 
Hi Tom,

This PR is to allow uploading files to Raster source in the raster source form. I think it's a good idea to have you reviewing it now while I am working on the upload queue.

Kr, Hoan

Derryrover:
@hoanphungt 
Hi Hoan,

I think code technically and functionally good.
My biggest concern is that we now upload all at once many .tiff files (after clicking save).
I previously agreed with Casper that we would upload the files one at a time (wait untill api call returns before sending next file).
Please check how it works in the existing management client.
@caspervdw Is it still a requirement that we upload the files one at a time? I believe it was for performance reasons on backend.

kr Tom

hoanphungt:
@Derryrover 
Hi Tom,

In this commit: https://github.com/nens/lizard-management-client/pull/517/commits/9dfdeb830c0d8e7c937871427055529938b10590, I updated the `sendDataToLizard` function to `sendDataToLizardRecursive` and to call it recursively. It's similar to the old way but I make it simpler. I think it functions well enough. What do you think?

Kr, Hoan

PR #527, opened 2020-11-23
Fix scrollbar in raster layer form

hoanphungt:


hoanphungt:
@Derryrover 
Hi Tom,

This solution seems working. Not very nice but it works for both the raster form and the alarm notification form. Could you have a look again? 

Thanks!

Derryrover:
@hoanphungt 
Looks good. Well done!

PR #529, opened 2020-11-24
Add upload queue

hoanphungt:


hoanphungt:
@Derryrover 
Hi Tom,

I drafted this PR to experiment with polling the tasks endpoint and showing the tasks status in the upload modal. Of course, the design would change but I would not expect a lot of work for them. Currently, I stored the tasks in the Redux store and only poll to the first task in the queue until it is done, and then go to the next one. 

The main disadvantage for now is that the Tasks state in the Redux store doesn't know how many tasks will be added and also can't tell when all the upload tasks are finished.

What do you think?

Kr, Hoan

Derryrover (approved):

src/reducers.js
```diff
@@ -300,19 +304,55 @@ function rasterSourceUUID(state = null, action) {
...
-      };
+    case ADD_FILES_TO_QUEUE:
+      return action.files.map(file => {
+        return {
+          "name": file.name,
+          "size": file.size,
+          "uuid": null,
+          "status": 'WAITING'
+        };
+      });
```
Derryrover, line 317: Could it be that this removes the existing files from the queue as well?

hoanphungt: Yes, currently, adding new files to the upload queue will remove the old ones. I thought the function `sendDataToLizardRecursive` will not work correctly in this case. However, I just changed the uploadFiles reducer and tested it. It worked like a charm. This is better than the old solution for sure. Thanks for your finding!

PR #538, opened 2020-12-07
draft delete confirm modal

Derryrover:


src/data_management/rasters/RasterLayerTable.tsx
```diff
@@ -165,6 +206,57 @@ export const RasterLayerTable = (props:any) =>  {
...
+            setDeleteFunction(null);
+          }}
+          disableButtons={busyDeleting}
+         >
+           
+           <p>Are you sure? You are deleting the following raster layers:</p>
+           
+           {modalDeleteContent(rowsToBeDeleted, busyDeleting)}
+           
+         </ConfirmModal>
```
hoanphungt, line 231: If there are many rasters selected for deletion (say 10) then the modal layout will be broken

src/data_management/rasters/RasterLayerTable.tsx
```diff
@@ -165,6 +206,57 @@ export const RasterLayerTable = (props:any) =>  {
...
+             setDeleteFunction(null);
+           }}
+           disableButtons={busyDeleting}
+         >
+           <p>Are you sure? You are deleting the following raster-layer:</p>
+           {modalDeleteContent([rowToBeDeleted], busyDeleting)}
+         </ConfirmModal>
+        :
+          null
+        }
```
hoanphungt, line 259: I think it is not necessary to treat deletion of 1 raster differently than deletion of multiple rasters. I guess there is technical reason that you implement the deletion separately for 1 raster and multiple rasters?

hoanphungt (changes requested):
Hi Tom,

Only comment that the modal is now broken if many layers are selected for deletion. Also, if you use ESC to close the modal then it will break the app.

Generally, I find there are way too many `any` type and `@ts-ignore` used. It makes little sense to use typescript in this case then. But you can deal with it later in a separate ticket. I also notice that recently you used a lot of in-line css styling in the tsx file which is not uniform with how we style our components across the app (using css modules). It might lead to extra work later on when we have to move all these styling to its own file.

Kr, Hoan

src/data_management/rasters/RasterLayerTable.tsx
```diff
@@ -10,141 +10,182 @@ import {ExplainSideColumn} from '../../components/ExplainSideColumn';
...
-        title={row.name}
-      >
-        <NavLink to={`${navigationUrlRasters}/${row.uuid}/`}>{row.name}</NavLink>
-      </span>,
-    orderingField: "name",
-  },
-  {
-    titleRenderFunction: () =>  "Based on",
-    renderFunction: (row: any) => 
+  const rasterSourceColumnDefenitions = [
```
hoanphungt, line 86: one more comment, I noticed the name here is `rasterSourceColumnDefinitions` while it should be rasterLayer instead.

src/data_management/wms_layers/WmsLayerTable.tsx
```diff
@@ -7,116 +7,173 @@ import wmsIcon from "../../images/wms@3x.svg";
...
-            triggerReloadWithBasePage={triggerReloadWithBasePage}
-            actions={[
-              {
-                displayValue: "delete",
-                actionFunction: deleteAction,
-              },
-            ]}
-          />
-      );
+  const rasterSourceColumnDefenitions = [
```
hoanphungt, line 90: Also here, the name is incorrect. It should be `wmsLayerColumnDefinitions` instead. Also spelling error here: `Definitions` instead of `Defenitions`. The same in other places.

src/data_management/rasters/RasterLayerTable.tsx
Derryrover: done

src/data_management/rasters/RasterLayerTable.tsx
Derryrover: Fixed this in latest commit.

src/data_management/rasters/RasterLayerTable.tsx
Derryrover: Yes this has technical reasons.
I also thought about this option. I think it is good to investigate, but too much work for now. I made this ticket:
https://github.com/nens/lizard-management-client/issues/556

src/data_management/wms_layers/WmsLayerTable.tsx
Derryrover: done

Derryrover:
@hoanphungt 

Hi Hoan,

Thanks for the review!
I took actions on most point as can seen in the comments.
I also fixed the escape button problem, by handling the "handleClose" property in Overlay component.

This leaves 2 open issues:
- inline css: I took action and factored out the content of the delete modal. So most inline css is now at one place. It is thus now a small problem. I right now do not see the need to remove the css from being inline because: some css is dynamic based on javascript and most css is layout css and this type is sometimes more clear inline (imo).

- the ts-ignores + any's: biggest problem with the current table solution is that a row has no type. Giving the row a type would require typing all different rows as they will come from the api. I think I might wait for a swagger definition, because typing them all is a lot of work. I made this ticket: https://github.com/nens/lizard-management-client/issues/558

kr Tom

PR #539, opened 2020-12-07
Update min max zoom field for wms layer form

hoanphungt:


hoanphungt:
@Derryrover 
Hi Tom,

I made some changes to the min max zoom fields. Please let me know what you think. 

Kr, Hoan

PR #553, opened 2020-12-17
attempt fix security vulnerabilities

Derryrover:


Derryrover:
@remcogerlich 

Hi Remco,
Could you review?
I tried to fix some security vulnarabilities for management client.
I could fix them all, but this introduces to new problems:
- postcss is only compatible with nodejs v12 (see readme)
- babel loader is different for react-script vs storybook so I skip preflight check to enable this (wee .env)

kr Tom

Derryrover:
@remcogerlich 

Hi Remco,
Time to review this one?
I can also merge without review.
kr Tom

PR #555, opened 2020-12-21
Scenario form

hoanphungt:


src/form/useForm.ts
```diff
@@ -1,10 +1,6 @@
 import React, { useState } from 'react';
-import { ColorMapOptions } from './ColorMapInput';
-import { SpatialBounds } from '../types/mapTypes';
-import { AcceptedFile } from './UploadRasterData';
-
-type Value = string | number | boolean | string[] | {} | ColorMapOptions | SpatialBounds | AcceptedFile[] | null |undefined;
 
+type Value = any;
```
hoanphungt, line 3: I would like to try a new approach here by giving value an any type. This removes the need to have `as string` or `as number`, ... everywhere in the form.

Derryrover: For now good.

Derryrover (approved):

PR #570, opened 2021-01-07
Tom colormap Modal on rasterlayer form

Derryrover:


src/form/ColorMapInput.tsx
```diff
@@ -235,6 +267,31 @@ const ColorMapInput: React.FC<ColorMapProps & InjectedIntlProps> = (props) => {
...
     >
+      {showCustomColormapModal? 
+        // <Modal
+        //   title={'CUSTOM COLORMAP'}
+        // >
+        <ModalBackground
+          title={'CUSTOM COLORMAP'}
+          handleClose={() => setShowCustomColormapModal(false)}
+          // value depends on content, but should be precisely hardcoded
+          height={'816px'}
```
hoanphungt, line 278: You can also give value in percentage (i.e. 60%), not only in pixels. There is a default value for the modal as well (70% width and height of the whole screen). So if you don't pass anything to width and height, it will be 70% width and height.

src/components/Checkbox.tsx
```diff
@@ -48,6 +50,7 @@ const Checkbox: React.FC<Props> = (props) => {
         onFocus={onFocus}
         onBlur={onBlur}
         readOnly={readOnly}
+        form={form}
```
hoanphungt, line 53: why do you need this form parameter?

src/data_management/rasters/RasterLayerForm.tsx
```diff
@@ -405,18 +423,21 @@ const RasterLayerForm: React.FC<Props & PropsFromDispatch & RouteComponentProps>
...
         <div
           className={formStyles.ButtonContainer}
         >
           <CancelButton
             url={'/data_management/rasters/layers'}
+            form={"raster_layer_form_id"}
           />
           <SubmitButton
             onClick={tryToSubmitForm}
+            form={"raster_layer_form_id"}
```
hoanphungt, line 437: Why do you need this form element everywhere? is it to distinguish between the raster layer form and the custom color map form?

src/form/CustomRadioSelect.tsx
```diff
@@ -0,0 +1,66 @@
...
+interface Props {
+  title: string | JSX.Element,
+  name: string,
+  value: string,
+  valueChanged: (event: any) => void,
+  readonly?: boolean
+  options: Option[]
+};
+
+export const CustomRadioSelect: React.FC<Props> = (props) => {
```
hoanphungt, line 19: This is a nice component to have. I think we can also reuse it for the access modifier.

src/form/TextInput.tsx
```diff
@@ -5,17 +5,21 @@ import formStyles from "../styles/Forms.module.css";
...
   onFocus?: (e: React.ChangeEvent<HTMLInputElement>) => void,
   onBlur?: () => void,
   handleEnter?: (e: any) => void,
   triedToSubmit?: boolean,
-  readOnly?: boolean
+  readOnly?: boolean,
+  form?: string,
+  type?: string,
+  required?: boolean,
+  showUpDownArrows?: boolean, // only works if type = "number"
```
hoanphungt, line 22: I think if we want to have the upDownArrows then it's better to use the default number input HTML element instead of customising it. It doesn't make sense to have the upDownArrows for a text input component.

src/data_management/rasters/RasterLayerForm.tsx
```diff
@@ -183,10 +185,14 @@ const RasterLayerForm: React.FC<Props & PropsFromDispatch & RouteComponentProps>
...
       <form
-        className={formStyles.Form}
+        // remove this class because it gives element height
+        // className={formStyles.Form}
         onSubmit={handleSubmit}
         onReset={handleReset}
+        id={"raster_layer_form_id"}
       >
+      </form>
+      <div className={formStyles.Form}>
```
hoanphungt, line 195: I am not sure if it is a good choice here. Do I understand correctly that you made a form that will handle the onSubmit event for the real form container (which is now a div)? Is it because otherwise it will get conflict with the color map form? I think in general the decision to use a form for the custom color map modal (inside the raster layer form) is costing us a lot now. It can become simpler if it is just a modal component that will output a value based on the inputs.

src/form/ColorMapInput.tsx
```diff
@@ -244,21 +301,41 @@ const ColorMapInput: React.FC<ColorMapProps & InjectedIntlProps> = (props) => {
...
-          onFocus={onFocus}
-          onBlur={onBlur}
-          readOnly={readOnly}
-        />
+        <div style={{position: "relative"}}>
+          <SelectBox
+            title={''}
+            choices={[["Custom colormap","Custom colormap","+ Create new colormap for this raster"],...colorMaps]}
+            value={JSON.stringify(colorMapValue.customColormap) !=="{}" && JSON.stringify(colorMapValue.options) ==="{}" ? "Custom colormap" : (colorMapType && colorMapType.colorMap) || null}
+            name={name + '_colorMapselect'}
```
hoanphungt, line 309: This will probably break the dynamic help text in the explain side column (it depends on the name of the field and this field name must be included in the `helpTextForForms.tsx` component)

hoanphungt (commented):
Looks good. I have some comments. I will add my comments in the related ticket on the board.

src/components/Checkbox.tsx
Derryrover: Hi Hoan,

I understand your concerns regarding the form element and all.
I should have documented the <form/> problem+solution within the code (done that below).

The workaround with the empty <form/> element is needed because otherwise I had to nest a <form/> inside a <form/> which is illegal html.
Therefore I used the 'workaround' based on the second answer here:
https://stackoverflow.com/questions/3430214/form-inside-a-form-is-that-alright?lq=1

This does complicate the situation, but I really like to have the form for the colormap. Why?
Well first of all the colormap needs validation before it should be written back to the raster-layer form.
Creating this validation from scratch is hard and will also not be consistent with or other validation (which all use the form). Only doing this validation once the user submits the rasterlayer, would require us to re-open the modal (bad imo).

Second of all I think the Colormap is actually a form. Not only will this keep or code more consistent, it is also semantically correct. I know the backend does not threath the colormap as its own entity, but I think this is a 'implementation detail' on their side and I try to prevent copying such details to our code.

Hope this convinces you or otherwise feel free to comment question.
__________________________________________________________________________
I added the following comment to the code:
{/* 
      I also use a form inside the colormap modal.
      A form inside a form is not valid html.
      Therefore I used the workaround here:
      https://stackoverflow.com/questions/3430214/form-inside-a-form-is-that-alright?lq=1
      Answer from user: "ilevent"
      */}

src/form/TextInput.tsx
Derryrover: I will for now not change this, but I think you do have a point.
Not sure about this honestly.

My motivation behind wanting to re-use this TextInput is to be able to re-use the following code.
Might be good if we only have that at one spot.

if (validated) {
        myInput.current.setCustomValidity('');
      } else {
        myInput.current.setCustomValidity(errorMessage || '');
      };

hoanphungt: Ah I just look at the code again. It's actually a good idea to use this here. It's just a bit confusing that the TextInput field now can sometimes become NumberInput that supports only number. It would be better to rename the field to TextNumberInput in my opinion. However, it's not important, we can just keep it the way it is now.

src/components/Checkbox.tsx
hoanphungt: Ah ok. I understand your points. I am still not super convinced that we used a form inside a form. There must be a reason that a form nested inside another form is invalid in HTML. But your points are also valid so lets keep it this way.

src/form/ColorMapInput.tsx
Derryrover: gave the colormap-select versus the min and max different id's and added those to helpTextForms.tsx

src/form/ColorMapInput.tsx
Derryrover: Gave the width and height in percentage

src/form/TextInput.tsx
```diff
@@ -60,19 +68,21 @@ export const TextInput: React.FC<MyProps> = (props) => {
...
           placeholder={placeholder}
           onChange={valueChanged}
-          onFocus={onFocus}
-          onBlur={onBlur}
-          value={value || ""}
+          value={value || (value===0? value: "")}
           onKeyUp={handleEnter}
           readOnly={!!readOnly}
-          // disabled={!!readOnly}
+          disabled={!!readOnly}
```
hoanphungt, line 78: @Derryrover 
Hi Tom, I missed this one in the last review. Is there a reason for you to switch this disabled parameter on? The reason I commented it out is because I want the focus attribute to work even when the field is read-only (for the help text in the side column). I didn't document it in the code but it's necessary to do it this way. If it is not going to interfere with your work then I will remove them again. Is it ok?

Derryrover: @hoanphungt
Sorry I already merged.
I will fix this now

hoanphungt: @Derryrover 
Ah don't worry about it! I fixed it in my PR: https://github.com/nens/lizard-management-client/pull/589

PR #578, opened 2021-01-18
fix url sso parameter

Derryrover:


Derryrover:
I have not heard anymore that this is still a problem.
Close it

PR #593, opened 2021-01-21
draft tile personal api keys

Derryrover:


src/App.js
```diff
@@ -302,6 +307,20 @@ class App extends Component {
...
+                    exact
+                    path={`/personal_api_keys/new`}
+                    component={NewPersonalApiKey}
+                  />
+                  <Route
+                    exact
+                    path="/personal_api_keys/:uuid"
+                    component={EditPersonalApiKey}
+                  />
+                </Switch>
```
hoanphungt, line 323: Nice that you are using Switch from react-router here

src/components/Checkbox.module.css
```diff
@@ -23,8 +23,6 @@
   pointer-events: none;
   visibility: hidden;
   position: absolute;
```
hoanphungt, line 25: I think since you are using flex-box to position the checkmark, we can safely remove the position relative and absolute from the CheckboxContainer and the checkbox.

src/components/TableStateContainer.tsx
```diff
@@ -225,6 +226,7 @@ const TableStateContainerElement: React.FC<Props> = ({ gridTemplateColumns, colu
           marginTop: "16px",
         }}
       >
+        { textSearchBox !== false?
```
hoanphungt, line 229: This condition is a bit strange. Since textSearchBox is a boolean, you should be able say: `textSearchBox ? <TableSearchBox /> : null`. I understand that in case textSearchBox is undefined, you still want to show the search box but it's a bit unclear to do it this way. It should be the other way around, if you don't specify the textSearchBox parameter then the component will not show the search box at all.

src/form/CheckBox.tsx
```diff
@@ -22,7 +25,10 @@ export const CheckBox: React.FC<CheckBoxProps> = (props) => {
     form,
     onFocus,
     onBlur,
-    readOnly
+    readOnly,
+    validated,
+    errorMessage,
+    htmlTitle,
```
hoanphungt, line 31: I wonder why you validated the form checkbox in the other Checkbox component, not in this component? This component directly belongs to the form why the other checkbox can used in other places. If you want to reuse the checkbox validation in other places outside of the form as well then it makes sense though.

src/personal_api_keys/NewPersonalApiKey.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  useEffect(() => {
+    setallPersonalApiKeysFetching("RETRIEVING");
+    fetch('/api/v4/personalapikeys')
+    .then(response=>{
+      if (response.status === 200) {
+        return response.json();
+      } else {
+          // todo fix this erro notification
+          // props.addNotification(status, 2000);
```
hoanphungt, line 19: do you have a ticket to fix this error notification?

src/personal_api_keys/PersonalApiKeyForm.tsx
```diff
@@ -0,0 +1,281 @@
...
+      .then(data => {
+        const status = data.status;
+        if (status === 201) {
+          // setting both modal and snackbar message is a bit too much so I comment snackbar message out
+          // props.addNotification('Success! Personal API key generated', 2000);
+          data.json().then((record: any)=>{
+            setApiKeyString(record.key+'');
+          })
+          // do not redirect because we need to show the api key modal
+          // props.history.push('/personal_api_keys/');
```
hoanphungt, line 100: I think commented out code like these can be safely deleted from the code base.

src/utils/helpTextForForms.tsx
```diff
@@ -168,4 +168,10 @@ export const lableTypeFormHelpText: HelpText = {
       <p><i>If you are not an administrator, this field is always pre-filled with the current organisation.</i></p>
     </>
   ),
+}
+
+export const personalApiKeysFormHelpText: HelpText = {
+  default: "Personal API keys can be used to authenticate external applications in Lizard",
+  name: "Name of the personal api key",
+  scope: "Defines what the personal api key can be used for",
```
hoanphungt, line 176: This help text for scope field cannot be shown right now because the field name is different.

src/App.js
```diff
@@ -302,6 +307,20 @@ class App extends Component {
                 <Route exact path="/" component={Home} />
                 <Route path="/alarms" component={AlarmsApp} />
                 <Route path="/data_management" component={DataManagementApp} />
+                
+                <Switch>
+                  <Route exact path="/personal_api_keys" component={PersonalApiKeysTable} />
```
hoanphungt, line 312: As Casper mentioned, this personal_api_keys shows up on the navigation tab as `Personal API_KEYS` with underscore.

hoanphungt (commented):
Looks good! only some small comments. I think it's viable to do it this way for low-risk and urgent features that need to be on staging early for testing purposes. Well done!

src/components/Checkbox.module.css
Derryrover: - position absolute cannot be removed from the checkmark, because then it will be rendered next to the input, instead on top of it. The flexbox centers the elements, but centers them together, so one becomes left in the center other right in the center. By giving the checkmark a position absolute it is removed from the flow, both elements are now centered in isolation from each other.
- The position relative can be removed from the container, although I rather not, because I do not understand why this works. The absolute specs say the element will be positioned absolute to its first relative (or absolute) parent. Maybe it works because relative is also the default. I am rather explicit here, I think that is more clear.

src/form/CheckBox.tsx
Derryrover: To do it in the same compomnent as where the  <input/> element is defined is easiest. It prevents me from needing to pas refs acros components.
Also I am at the moment not sure if any user input that can be invalid should ever be requested outside a form element.
So this means I could use the checkbox outside a form, but only if a invalid state is not possible (for example like in the list, where any checkbox being toggled on/off is always valid)

src/personal_api_keys/NewPersonalApiKey.tsx
Derryrover: Thanks for reminding me, I created this one:
https://github.com/nens/lizard-management-client/issues/598

src/personal_api_keys/PersonalApiKeyForm.tsx
Derryrover: done

src/utils/helpTextForForms.tsx
Derryrover: fixed here:
https://github.com/nens/lizard-management-client/pull/596

src/App.js
Derryrover: done here:
https://github.com/nens/lizard-management-client/pull/596

src/components/TableStateContainer.tsx
Derryrover: True, good point. improved the default here:
#598

PR #599, opened 2021-01-25
Exclude 3Di scenarios from raster layer and raster source tables

hoanphungt:


src/components/TableStateContainer.tsx
```diff
@@ -23,9 +23,10 @@ interface Props {
   textSearchBox?: boolean; // default true
   newItemOnClick: () => void | null;
   queryCheckBox: {text: string, adaptUrlFunction: (url:string)=>string} | null;
+  scenarioIsNull?: boolean;
```
Derryrover, line 26: @hoanphungt 
Hi Hoan,
I am not sure about this change.
Although this works,  it does not scale up well.
What if in another table we need another default url parameter? We need to keep adding props.
I think url parameters that are standard for a certain table should either be appended as string to "baseUrl" or be passed as a newly to be made props called something like "defaultUrlParameters". 
Please let me know what you think
kr Tom

hoanphungt: Ah you are right. I will improve it this way.

hoanphungt:
@Derryrover 
Could you review it again?

Derryrover (approved):
Excellent !

PR #600, opened 2021-01-26
Style checkbox and action buton to center position

hoanphungt:


hoanphungt:
Center position of checkbox and actions button is not needed.

PR #615, opened 2021-01-29
Delete raw button only if scenario has raw data

hoanphungt:


tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es5",
+    "target": "es6",
```
hoanphungt, line 3: I changed the ts config compiler options to "ES6" so the typescript does not complain about this method: `[...new Set([...checkBoxes ,...allCurrentPageUuids])]`

src/components/TableStateContainer.tsx
```diff
@@ -5,31 +5,35 @@ import {ColumnDefinition} from './Table';
...
 import {DataRetrievalState} from '../types/retrievingDataTypes';
 import unorderedIcon from "../images/list_order_icon_unordered.svg";
 import orderedIcon from "../images/list_order_icon_ordered.svg";
 import styles from './Table.module.css';
 import buttonStyles from '../styles/Buttons.module.css';
 
+interface checkboxAction {
+  displayValue: string,
+  actionFunction: Function,
+  nameOfDependentDataField?: string,
```
hoanphungt, line 19: I am thinking of another name for this variable, for example: `queryDataField`. What do you think?

Derryrover: Hi Hoan,
I think this is done almost right.
I think a more scalable improvement might be that "nameOfDependentDataField" would instead be a function that takes the whole table row and returns a boolean. It could be called something like "testIfApplicableFunction".
For the rest i think this looks all very good!
kr Tom

Derryrover (changes requested):
see comment

src/components/TableStateContainer.tsx
hoanphungt: Thanks for your comment! Sounds like a very scalable improvement. I added this function to the `checkboxActions` of the ScenarioTable:

`checkIfActionIsApplicable: (row: any) => row.has_raw_results === true`

PR #628, opened 2021-02-03
Add new Select dropdown

hoanphungt:


src/form/Dropdown.tsx
```diff
@@ -7,7 +7,7 @@ interface MyProps {
   name: string,
   value: string,
   options: string[],
-  valueChanged: (e: React.ChangeEvent<HTMLInputElement>) => void,
+  valueChanged: (value: string | null) => void,
```
hoanphungt, line 10: This Dropdown.tsx does not require a review. This is an attempt to build the input dropdown but I have not finished it yet. I might come back to it later.

src/stories/Form.tsx
```diff
@@ -0,0 +1,140 @@
+import React, { useEffect, useState } from 'react';
```
hoanphungt, line 1: No need to review this as it's for demo purpose only

Derryrover (approved):

PR #635, opened 2021-02-08
prevent colormapinput modal explainbox trigger

Derryrover:


src/form/ColorMapInput.tsx
```diff
@@ -115,6 +117,11 @@ const ColorMapInput: React.FC<ColorMapProps & InjectedIntlProps> = (props) => {
   });
   const [showCustomColormapModal, setShowCustomColormapModal] = useState(false);
```
hoanphungt, line 118: Since you kept the `showColomapModal` in the raster layer form, it's unnecessary to also create a new one here in the ColormapInput component. Then you don't have to keep them in sync with each other in the below function.

src/form/ColorMapInput.tsx
```diff
@@ -115,6 +117,11 @@ const ColorMapInput: React.FC<ColorMapProps & InjectedIntlProps> = (props) => {
...
 
+  const setShowCustomColormapModalWithCallback = (bool:boolean) => {
+    setShowCustomColormapModalProp && setShowCustomColormapModalProp(bool);
+    setShowCustomColormapModal(bool);
+  };
+
```
hoanphungt, line 124: This function is unnecessary if you keep the boolean in the parent component only.

hoanphungt (changes requested):
I think it's better to keep the showColormapModal variable in the parent component only and remove the one in the child component.

hoanphungt (approved):

src/form/ColorMapInput.tsx
```diff
@@ -270,129 +270,126 @@ const ColorMapInput: React.FC<ColorMapProps & InjectedIntlProps> = (props) => {
   const placeholderMaximumColorRange = intl.formatMessage({ id: "placeholder_maximum_color_range" })
 
   return (
-    <label
-      htmlFor={name}
```
hoanphungt, line 267: I think a simpler solution is to remove the `htmlFor={name}` attribute of the label here instead of removing the whole label tag (it has some specific styling for the label namely `marginBottom` which is now totally removed)

src/form/ColorMapInput.tsx
```diff
@@ -263,129 +266,126 @@ const ColorMapInput: React.FC<ColorMapProps & InjectedIntlProps> = (props) => {
...
+                triedToSubmit={triedToSubmit}
+                valueChanged={(colormap)=>{ 
+                  colorMapChanged(colormap); 
+                }}
+                placeholder={placeholderColorMapSelection}
+                showSearchField={true}
+                readOnly={readOnly}
+                form={form}
+                onFocus={onFocus}
+                onBlur={onBlur}
```
hoanphungt, line 324: I think another solution is to remove the `onBlur` event here. I tried it and it seems it's a good enough solution instead of moving the a part of the colormapInput state to the raster layer form.

src/form/ColorMapInput.tsx
Derryrover: removing the htmlFor={name} did not resolve the issue on firefox. Clicking anywhere on the colormap input element still opened the dropdown. Didnot test for chrome.
I will try to fix the issue with marginBottom

hoanphungt: Strange. It worked on my firefox. Could you try to give it a different name? such as `htmlFor={name+'colorMapInput'}`. The problem is there is a conflict here. The `SelectBox` also has a lable with `htmlFor={name}`, so if they don't have the same name then it would not trigger the dropdown under the color map input.

Derryrover:
closed, merged this one instead:
nens/lizard-management-client#639

PR #637, opened 2021-02-09
tiledata

Derryrover:


src/home/Routes.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <Route exact path="/alarms/notifications/:id" component={EditNotificationApp} />
+        <Route exact path="/alarms/groups" component={AlarmGroupsApp} />
+        <Route exact path="/alarms/groups/new" component={NewAlarmGroupApp} />
+        <Route exact path="/alarms/groups/:id" component={AlarmGroupsDetail} />
+        <Route exact path="/alarms/contacts" component={AlarmContactApp} />
+        <Route exact path="/alarms/contacts/new" component={NewContactApp} />
+        <Route exact path="/alarms/contacts/:id" component={ContactDetail} />
+        <Route exact path="/alarms/templates" component={AlarmTemplatesApp} />
+        <Route exact path="/alarms/templates/new" component={NewTemplateApp} />
+        <Route exact path="/alarms/templates/:id" component={AlarmTemplatesDetail} />
```
hoanphungt, line 82: I thought the map function already rendered all the Routes (but without the component). Is it a duplication of what is returned from the map function?

hoanphungt (commented):
Looks good! I think it's nice that all app tiles configuration is now in one place and this becomes quite backend configurable later on.

src/home/Routes.tsx
Derryrover: The map function renders all routes that show the 'home' component.
The home components renders the tile based layout.

The hardcoded routes below are all for rendering no-tile based layouts (tables, forms etcetera)

remcogerlich: I feel the map function should render a routing component for that route. That can then have its own router that chooses which part to render, home, table, form et cetera. Use nested routing.

Derryrover: I think nested routing could be good, but I have not done it for the following 2 reasons:
- It will be harder to generate the routing from a list of configured tiles (assuming that the the list of tiles is itself a flat list and thus not a json tree structure)
- It will require more knowledge from the routing library. I would need to investigate.

My plan was to keep the list of tiles a flat list with its own database table that a user can then configure on the management page. Therefore I prefer not to use a nested structure. Is this acceptable for you? Some changes to the url will require a bit more work, but nothing too complicated I guess.

remcogerlich: - The question is to either hardcode this list here, or inside components. It's hard coded both ways.

- A user would probably configure on the level of the appTiles list, and not on individual parts (no users wants to configure that he wants to have the alarms, but not that one form on the alarms subpages). Same general reason why you configured some of them in the appTiles list and not all of them. So it wouldn't be touched by more configuration by the user.

- Nesting with hooks is very simple ( https://reactrouter.com/web/example/nesting ) but I don't know how it was with classes exactly.

PR #638, opened 2021-02-09
Dropdown component for action button

hoanphungt:


src/components/ActionList.module.css
```diff
@@ -0,0 +1,32 @@
...
+}
+.ActionList > li:hover {
+  background-color: #DEEBFF;
+}
+
+/* CSS to render ActionList dropdown on top of the last ActionButton in Table */
+.LastItemInTable .ActionList {
+  bottom: 100%;
+  top: unset;
+}
```
hoanphungt, line 32: The last ActionButton in the table requires the dropdown to become a dropup, otherwise users would need to scroll a bit more downward to see the dropdown actions.

src/components/Table.tsx
```diff
@@ -50,7 +51,10 @@ const Table: React.FC<Props> = ({tableData, setTableData, gridTemplateColumns, c
               return (
                 <React.Fragment key={idx}>
                   {columnDefinitions.map((definition, i)=>
-                    <span className={rowIsSelected? styles.Selected: styles.NotSelected} key={idx + i}>
+                    <span
+                      className={`${rowIsSelected ? styles.Selected : ''} ${tableData.length > 1 && idx === (tableData.length -1) && i === (columnDefinitions.length - 1) ? actionListStyles.LastItemInTable : ''}`}
```
hoanphungt, line 55: give classname for the last span of the table to change the dropdown element to dropup. For table with only 1 row then it would still stay as a dropdown.

Derryrover (approved):

PR #639, opened 2021-02-09
Remove onBlur in color map input and fix label name

hoanphungt:


hoanphungt:
@Derryrover 
Hi Tom,

For the bug which text in explain column changes when click around the custom color map modal, I feel like your fix introduces too many changes (in which includes the state change of the color map input). I made this PR in an attempt to solve it in a simpler way. Could you have a look and let me know what you think?

It solved the bug the explain box text change and the unexpected open/close of color map input dropdown. I tested it on both firefox and chrome and both worked correctly.

Kr, Hoan

PR #678, opened 2021-02-22
Add new template form

hoanphungt:


src/alarms/alarmtemplates/TemplateForm.tsx
```diff
@@ -0,0 +1,346 @@
...
+    message: ''
+  };
+
+  const onSubmit = (values: Values) => {
+    const body = {
+      name: values.name,
+      subject: values.subject,
+      text: values.message,
+      html: values.message
+    };
```
hoanphungt, line 113: here I still need to update both `text` and `html` field because the `text` field is required by backend. If we only send `html` when creating a email template then it will response with a 400 error code because of missing `text` field.

Derryrover (approved):

PR #681, opened 2021-02-23
guard selected organisation

Derryrover:


hoanphungt (commented):
Code looks good. Just one strange thing is that in development, I can use `yarn start` without providing my username and password and it still opens the management page with my login details.

Issue #680, opened 2021-02-22
Linking to page within management screens sometimes leads to white page


lexvand:
For instance: https://hhnk.lizard.net/management/#/data_management/scenarios/

![image.png](https://images.zenhubusercontent.com/5ede3192cc5c2c4503d452e7/4a8dd53b-fe06-440f-87e2-b46a285994db)

Refresh (both F5 and hard refresh) does not help. It doesn't request for the user roles, hence the error in the console. I guess the internal authorisation method doesn't work correctly. Navigating from the main management to the same location does work.

EDIT: Could also be that something goes wrong with redirecting to the login page when the user is not logged in. Tom suggests that it is caused by error: TypeError: this.props.selectedOrganisation is null

**Testing**
This bug is 100% reproducable by going directly to the following url in a new incognitio window:
https://nxt3.staging.lizard.net/management/#/data_management/scenarios
https://hhnk.lizard.net/management/#/data_management/scenarios/
Do not login first or go to the main page first.
The bug is that these wil show a blank page.
The desired behaviour is that they will redirect to the login page.

lexvand:
@lisannewapstra Eerdere issue stond in verkeerde repo.

Derryrover:
wip pr
https://github.com/nens/lizard-management-client/pull/681

hoanphungt:
@Derryrover 
Code looks good. However, in dev, I experienced some strange things:

1. I can start the app using `yarn start` without having to provide `PROXY_USERNAME` and `PROXY_PASSWORD`. However, the app still used my login details for bootstrap.
2. If I open the page in incognito mode, using `yarn start` leads me to the login page. After login, I am redirected to the lizard portal and I cannot go to the management page anymore (going to `localhost:3000` or `http://localhost:3000/management#/` always redirected to the lizard portal). 

I am not sure if these problem will still happen if we go to staging.

Kr, Hoan

Derryrover:
@hoanphungt 

Hi Hoan Thanks for the review
I merged this.
The problem you describe is known to me.
I think we have the following situation:

- If you not supply username password, the browser assign the authentication cooky / token to the url of staging somehow so the backend still thinks we are logged in. I am not sure why the browser assigns this auth cookie / token, because the domain is different (localhost vs nxt3.staging.lizarn.net).
Anyway, I think this is a problem on dev not on staging. Let us see what happens on staging.
[...]

PR #687, opened 2021-02-24
fix storage in default scenario helptext

Derryrover:


src/reducers.js
```diff
@@ -100,6 +102,29 @@ function organisations(
...
+function usage (
+  state = {
+    raster_count: 0,
+    raster_total_size: 0,
+    scenario_count: 0,
+    scenario_total_size: 0,
+    timeseries_count: 0,
+    timeseries_total_size: 0,
+    isFetching: false,
+    timesFetched: 0,
```
hoanphungt, line 114: Is there a reason to use both snake cases and camel cases here? I guess because data comes from API is in snake case.

src/data_management/rasters/RasterSourceTable.tsx
```diff
@@ -25,6 +29,8 @@ export const RasterSourceTable = (props:any) =>  {
   const [deleteFunction, setDeleteFunction] = useState<null | Function>(null);
   const [busyDeleting, setBusyDeleting] = useState<boolean>(false);
   const [showDeleteFailedModal, setShowDeleteFailedModal] = useState<boolean>(false);
+  const scenarioTotalSize = useSelector(getScenarioTotalSize);
```
hoanphungt, line 32: I think you should get `rasterTotalSize` here instead of `scenarioTotalSize`

src/actions.js
```diff
@@ -86,7 +87,7 @@ export function fetchOrganisations() {
     fetch(url, opts)
       .then(responseObj => responseObj.json())
       .then(responseData => {
-        const organisation = state.organisations.selected;
+        const selectedOrganisationLocalStorage = getLocalStorage("lizard-management-current-organisation", null);
```
hoanphungt, line 90: I don't get why you need to get the selected organisation from the local storage and how is this change related to this issue to show the storage of rasters and scenarios. Does it fix have anything to do with this bug: https://github.com/nens/lizard-management-client/issues/680?

src/components/ExplainSideColumn.tsx
```diff
@@ -11,6 +11,23 @@ interface MyProps {
...
+    if (typeof value === "object" && value !== null) {
+      if (seen.has(value)) {
+        return;
+      }
+      seen.add(value);
+    }
+    return value;
+  };
+};
+
```
hoanphungt, line 30: I think this is because you are currently saving the selected organisation in the local storage as a string and then later on, you try to stringify this string again and it caused a circle. I think it could be solved by not storing the selected organisation in local storage as JSON string but as an object.

hoanphungt: I just checked and it's not possible to store an object in local storage. So it means you need to parse it to an object first before using it.

src/actions.js
hoanphungt: Also, this leads to bug. If I don't have any selected organisation in the local storage, it will cause error: `Cannot read property 'roles' of null`. So overall, I think this is not a good change.

hoanphungt (commented):
Comments in PR

src/components/ExplainSideColumn.tsx
Derryrover: It is needed for the line below that uses a jsx component as key:
key={JSON.stringify(explanationText, getCircularReplacer)}
We could also adapt the JSX element manually to a different non circular object.
Anyway I suspect that introduces new surface for bugs, so I prefer this patch.

src/data_management/rasters/RasterSourceTable.tsx
Derryrover: fixed

src/reducers.js
Derryrover: Yes because it comes from the api. I choose this approach of changing as least as possible to the original data structure.

src/actions.js
Derryrover: You are right!
I need to look into this.
I will start to merge the fix for the 680 bug to see if it helps.

src/reducers.js
```diff
@@ -60,7 +61,7 @@ function organisations(
     timesFetched: 0,
     available: [],
     availableForRasterSharedWith: [],
-    selected: getLocalStorage("lizard-management-current-organisation", null),
+    selected: null,
```
Derryrover, line 64: I could remove "getLocalStorage("lizard-management-current-organisation" here because I dispatch it in the action instead.
Dispatching it in the action allows me to also trigger a fetch to the usage api on startup.

src/actions.js
Derryrover: @hoanphungt 
Hi Hoan,
I reproduced your error and believe I fixed it.
Also in the meanwhile I found bugs with the routing that I also fixed (I think so).
The "getLocalStorage("lizard-management-current-organisation", null); " here is needed to retrieve the selected organisation that the user selected during last usage of the app.
Previously this was done in the reducer default value, but here it is easier to also trigger the 'usage' api on startup. This usage api is fetched anytime a selected organisation is changed through the action. (but not if it is the default of the redux store).

hoanphungt (approved):
I think it looks good but needs to be tested on staging to see if these changes work correctly. Anyway, I think it would be better if you fix the routing issue and selected organisation bug in a separate issue/PR.

One more point, when I test it, the animation effect of the explain box is gone (so the border color of the explain box is not changing anymore when switching to another field). I will make a commit to improve this part a bit.

PR #691, opened 2021-02-26
Add raster alarm form

hoanphungt:


src/components/RasterPreview.js
```diff
@@ -2,134 +2,171 @@
...
+        onClick={handleMapClick}
+        className={styles.MapStyle}
+        zoomControl={false}
+        {...mapLocation}
+      >
+        <ZoomControl position="bottomright"/>
+        <TileLayer
+          url={`https://api.mapbox.com/styles/v1/nelenschuurmans/ck8sgpk8h25ql1io2ccnueuj6/tiles/256/{z}/{x}/{y}@2x?access_token=${mapBoxAccesToken}`}
+          attribution="&copy; <a href=&quot;http://osm.org/copyright&quot;>OpenStreetMap</a> contributors"
+          setLocation={setLocation}
```
Derryrover, line 150: I do not understand this setLocation.
Maybe I am missing something, but I can not find it as part of the api here:
https://react-leaflet.js.org/docs/api-components

src/components/RasterPreview.js
```diff
@@ -2,134 +2,171 @@
...
+            url="/wms/"
+            // @ts-ignore
+            styles={formatWMSStyles(raster.options.styles)}
+            layers={formatWMSLayers(raster.wms_info.layer)}
+            opacity={0.9}
+          />) : null}
+        <TileLayer
+          url={`https://api.mapbox.com/styles/v1/nelenschuurmans/ck8sgpk8h25ql1io2ccnueuj6/tiles/256/{z}/{x}/{y}@2x?access_token=${mapBoxAccesToken}`}
+          attribution="&copy; <a href=&quot;http://osm.org/copyright&quot;>OpenStreetMap</a> contributors"
+        />
```
Derryrover, line 163: I do not understand why a second tileLayer is needed here.
Could you explain it with a comment?

Derryrover (changes requested):
See 2 comments. For the rest looks good

src/components/RasterPreview.js
hoanphungt: Ah you are right. I didn't notice it. Kept it from the old code. It didn't do anything so I removed it. Also removed the attribution since it's not correct that we are using OpenStreetMap.

src/components/RasterPreview.js
hoanphungt: Also kept from the old code. It's a duplicate of the previous tile layer I guess so I removed it too.

PR #693, opened 2021-03-01
draft personal api key

Derryrover:


hoanphungt (commented):
Nice! README needs to be updated with how to change permission for the `startauth` file on ubuntu in order to run the file.

PR #717, opened 2021-03-16
Other try for authentication based on env variables

remcogerlich:


src/setupProxy.js
```diff
@@ -1,30 +1,36 @@
 const { createProxyMiddleware } = require('http-proxy-middleware');
 
 module.exports = function(app) {
-  const password = process.env.PROXY_PASSWORD;
-  const username = process.env.PROXY_USERNAME;
+  const env_prefix = process.env.PROXY_PREFIX || 'STAGING';
+  const url = process.env[env_prefix + '_URL'] || 'https://nens.lizard.net/';
```
Derryrover, line 5: This env_prefix makes things more complicated I think.
I understand it allows users to specify more PROXY_PREFIX variables in their ~./profile file.
But also in order to do that they would need to read and understand this code first

remcogerlich: They can ignore it, there is a default. It's purely optional. There is also explanation in the README that should be clear enough without having to look at the code. Isn't it? Is there something that can be improved about that?

What *is* daft is that I see that I made 'STAGING' the default prefix, but 'nens.lizard.net' the default domain. That doesn't work and needs to be fixed. Which one should be default, staging or production?

Ideally it should do the right thing with 1 setting, for the api key. And the rest is optional.

startauth.sh.example
```diff
@@ -0,0 +1,11 @@
+# at all times have only one line uncommented
+# PROXY_PREFIX value defines the name of the other 2 environmnet variables
+
+# example to proxy to staging
+PROXY_PREFIX=STAGING STAGING_URL=https://nxt3.staging.lizard.net/ STAGING_API_KEY=123456789STAGINGKEY yarn start
+
+# example to proxy to production
+# PROXY_PREFIX=LIZARD LIZARD_URL=https://demo.lizard.net/ LIZARD_API_KEY=123456789PRODUCTIONKEY yarn start
+
```
remcogerlich, line 9: Misschien moet de default prefix geen "STAGING" zijn, maar "PROXY" ?

Want op een command line een PREFIX meegeven is heel gek, je kunt ook gewoon de default variabele namen gebruiken dan. Maar `STAGING_URL=https://demo.lizard.net/` staat ook gek.

Als we de default prefix PROXY maken kan dit gewoon

`PROXY_URL=https://demo.lizard.net/ PROXY_API_KEY=123345667467 yarn start`

zijn.

PREFIX heeft alleen zin in de use case waar je een config file hebt met meerdere verschillende settings er in. Als je zoals hier het systeem "always leave one line uncommented" gebruikt dan zou je geen prefix moeten gebruiken, dat is dubbel.

Derryrover:
@remcogerlich 

Hi Remco,
Thanks for your help with this one.
I merged it.
If you see other improvements we can open a new branch

kr Tom

PR #720, opened 2021-03-18
Timeseries alarm form

hoanphungt:


hoanphungt:
@Derryrover 
Hi Tom,

In this PR, I added the TimeseriesAlarmForm which is pretty much similar to the RasterAlarmForm. The only field that has been changed is the TimeseriesSelection field instead of the RasterPointSelection. Could you review this `TimeseriesSelection` component?

Apart from the TimeseriesAlarmForm, I also cleaned up a lot of unused components that we used for the old management forms (in the `forms` directory and other components in the `components` directory). In my opinion, these components are outdated, not used anywhere in the current app so therefore not good to keep them in the code base. We can always go back to them if we need anything using Git version control. Please let me know what you think about this part and check if everything is still compiled correctly for you. I can of course revert all these changes if you deem it is still necessary to keep them.

Kr, Hoan

PR #725, opened 2021-03-22
Edit action for all tables

hoanphungt:


src/data_management/labels/LabeltypesTable.tsx
```diff
@@ -22,7 +22,7 @@ const columnDefinitions = [
         className={tableStyles.CellEllipsis}
         title={row.name}
       >
-        <NavLink to={`${navigationUrl}/${row.uuid}/`}>{!row.name? "(empty name)" : row.name }</NavLink>
+        <NavLink to={`${navigationUrl}/${row.uuid}`}>{!row.name? "(empty name)" : row.name }</NavLink>
```
hoanphungt, line 25: Removed the "/" at the end of the URL to remove the symbol ">" at the end of the navigation bar

src/components/TableActionButtons.tsx
```diff
@@ -16,19 +17,30 @@ interface Props {
   setTableData:any;
   triggerReloadWithCurrentPage:any; 
   triggerReloadWithBasePage:any;
+  navigationUrl: string;
```
Derryrover, line 20: Hi Hoan,
I think the navigationUrl is a good idea.
I would personally not have used it because I think it is more consistent to threat all actions 'equal' and have the action button not directly tied to the tableview.
Anyway, it does remove some code duplication, so maybe actually good.

2 points for possible improvement:
- navigationUrl seems to me not a good name. Maybe 'editUrl' ?
- Although edit is available for most or all tables I think it will eventually not be available for all tables. I think it should not be a mandatory action to have. Some tables might have no form.

kr Tom

hoanphungt: One more point is that in order to execute the EDIT action, we need to wrap the parent component inside a `withRouter` component in order to use the `props.history.push()`. So doing it in the TableActionButtons component helps to only do this once.

Good points that the naming is not good and the edit action should be made optional.

PR #747, opened 2021-03-25
Tom colormap colorpicker

Derryrover:


src/form/ColormapAllSteps.tsx
```diff
@@ -115,9 +115,36 @@ export const ColormapAllSteps: React.FC<Props> = (props) => {
...
+                  following div should act as a button.
+                  Button was turned into div, because it needs nested divs for the colorpicker.
+                  These divs need to be nested because clicking on them need to keep this div in focus 
+                  The div now namely uses a onblur to close the colorpicker (as replacement of the overlay with onclick previously used)
+                */}
+                <div 
+                  // this tabIndex is needed to make sure the div (that should actually be a button, but was made div for above reasons) does receive onfocus and onblur event
+                  tabIndex={0}
+                  onBlur={(event)=> {
+                    console.log('event', event)
```
hoanphungt, line 128: Remove console.log here

src/form/ColormapAllSteps.tsx
```diff
@@ -115,9 +115,36 @@ export const ColormapAllSteps: React.FC<Props> = (props) => {
...
                 style={{position: "relative"}}
               >
-                <button 
-                  onClick={()=> {if (visiblepickerIndex === ind) {setVisiblepickerIndex(null)} else {setVisiblepickerIndex(ind)} }}
-                  type="button"
+                {/* 
+                  following div should act as a button.
+                  Button was turned into div, because it needs nested divs for the colorpicker.
+                  These divs need to be nested because clicking on them need to keep this div in focus 
+                  The div now namely uses a onblur to close the colorpicker (as replacement of the overlay with onclick previously used)
```
hoanphungt, line 122: In my opinion, I think that you don't need to keep this div on focus all the time while clicking on the color map picker and it would look nicer if you don't show the outline at all (so `outline: none` for the button).  In this case, you can still use the button element instead of this div. Anyway, it's just a matter of preference, your solution is also working well.

src/form/ColormapAllSteps.tsx
```diff
@@ -115,9 +115,36 @@ export const ColormapAllSteps: React.FC<Props> = (props) => {
...
+                  onBlur={(event)=> {
+                    console.log('event', event)
+                    setVisiblepickerIndex(null);
+                  }}
+                  // Todo:
+                  // I added the keypress so user can also tab enter open this colorpicker.
+                  // anyway now the user might expect that excape closes it.
+                  // Not only does this not work (see next line, does not distinct between keypresses), 
+                  // also the event listener on <ModalBackground/> closes the entire modal on escape
+                  onKeyPress={()=>setVisiblepickerIndex(ind)}
```
hoanphungt, line 136: I think this added feature of using enter to open the color picker is not necessary. You still have to use your mouse to pick a color anyway. And you can use ENTER to open the color picker, but cannot close it using either 'Enter' nor 'ESC' so in the end, user is left to use the mouse anyway to use the color picker. I also noticed that the current version on staging works slightly better as you can use ENTER to open or close the color picker modal (not with ESC though).

hoanphungt (commented):
The issue is solved. I have some comments though.

src/form/ColormapAllSteps.tsx
Derryrover: commented it out for now

src/form/ColormapAllSteps.tsx
Derryrover: The focus is needed for the reason that a onblur closes tho colorpicker.
So clicking outside the colorpicker closes it.
Previously clicking outside the colorpicker also closed it, but this was done with an extra div overlay, blocking all other content. That old solution became difficult to accomplish, because i had to change some of the layout (the layout bug that this ticket fixes). Making the overlay div work correctly with the scroll is harder then using div that is in focus so I chose the latter.

PR #756, opened 2021-03-29
add form buttons

Derryrover:


src/data_management/timeseries/locations/EditLocation.tsx
```diff
@@ -20,16 +23,43 @@ const EditLocation = (props: RouteProps & RouteComponentProps) => {
...
 
+  useEffect (() => {
+    
+      (async () => {
+          // Todo: how to fix this error? I already check if currentRecord === null
+          // @ts-ignore
+          if (relatedAssetRequired && currentRecord && currentRecord !== null && currentRecord !== undefined && currentRecord.object) {
+            // Todo: how to fix this error? I already check if currentRecord === null
+            // @ts-ignore
+            const relatedAsset = await fetch(`/api/v3/${currentRecord.object.type}s/${currentRecord.object.id}/`, {
```
hoanphungt, line 44: This is because you are missing a type definition for currentRecord when you define it:

for example: `const [currentRecord, setCurrentRecord] = useState(null)` should become `useState<any>(null)` (use any or make type definition for it). In your case, currentRecord can only be null.

The same is applied to relatedAsset and relatedAssetRequired, there is no type definition for them.

src/data_management/timeseries/locations/EditLocation.tsx
```diff
@@ -12,6 +12,9 @@ interface RouteProps {
 const EditLocation = (props: RouteProps & RouteComponentProps) => {
   const organisations = useSelector(getOrganisations);
   const [currentRecord, setCurrentRecord] = useState(null);
+  const [relatedAssetRequired, setRelatedAssetRequired] = useState(true);
+  const [relatedAsset, setRelatedAsset] = useState(null);
+  // Todo fix this ts ignore
   // @ts-ignore
   const { uuid } = props.match.params;
```
hoanphungt, line 19: if you go to for example: EditTimeseries.tsx component, I added a type definition for RouteProps which is wrapped inside RouteComponentProps type: `RouteComponentProps<RouteProps>`. In the RouteProps, you can define the type for `props.match.params`.

hoanphungt (commented):
I am midway reviewing this. Will continue tomorrow.

src/form/GeometryField.tsx
```diff
@@ -0,0 +1,141 @@
...
+    onBlur,
+    handleEnter,
+    clearInput,
+    errorMessage,
+    triedToSubmit,
+    readOnly
+  } = props;
+
+  const [lat, setLat] = useState(NaN);
+  const [lng, setLng] = useState(NaN);
```
hoanphungt, line 44: missing type definition for lat and lng

hoanphungt: and any reason to use NaN instead of null as initial value here?

src/form/GeometryField.tsx
```diff
@@ -0,0 +1,141 @@
...
+      setLat(value.asset.value.view[0]);
+      setLng(value.asset.value.view[1])
+    } else if (value.location) {
+      setLat(value.location.lat);
+      setLng(value.location.lng); 
+    } else {
+      setLat(NaN);
+      setLng(NaN);
+    }
+  }, [value.asset, value.location]);
```
hoanphungt, line 58: I am not sure if this field can be reused by the RasterAlarmForm? Maybe with some tweaks it would be possible.

src/form/MapSelectAssetOrPoint.tsx
```diff
@@ -0,0 +1,246 @@
+// Todo: This component should probably also be used for the raster alarms
```
hoanphungt, line 1: It would be nice if this component can also be reused by the RasterPointSelection field in the Raster Alarms form. So the next step, we can reuse both the GeometryField and this MapSelectAssetOrPoint field for the RasterAlarmForm. I will make a separate ticket for this task.

hoanphungt (approved):

src/data_management/timeseries/locations/EditLocation.tsx
Derryrover: thanks !

src/form/GeometryField.tsx
Derryrover: Using "null / undefined / {} / other empty values" is always a difficult thing for me.
I thought to use NaN here because it indicates that the type is a number.
Anyway, I have no standard way of doing this.
If you have a reason to rather use null then NaN I am also fine.

PR #766, opened 2021-04-01
Set timeseries accessibility modal

hoanphungt:


hoanphungt:
@Derryrover 
Hi Tom,

I am working on the timeseries accessibility modal right now and I am currently having trouble to trigger the table to reload to show the updated accessibility. Now I need to refresh the page with F5 to see that the access modifier of the timeseries is indeed updated. 

I see that you pass the triggerReloadWithCurrentPage from the TableStateContainer to the deleteAction in order to achieve this page reloading. However, I prefer to be able to trigger this function within the AuthorisationModal.tsx component. Would you have some time to take a look at this PR and see if you would come out with some ideas?

Thanks in advance!

Derryrover (approved):
Good!

PR #769, opened 2021-04-02
Refactor delete actions

hoanphungt:


hoanphungt:
@Derryrover 
Hi Tom,

I made this PR to refactor the Delete Action. Currently, this piece of code to delete an object/multiple objects are duplicated in different tables and forms. So I made a DeleteModal component to reuse this piece of code. The deleteFunction is also refactored. Previously, the delete action for a single object and for multiple objects are separated by two different functions. I also combined them into one delete action.

I think it's a good improvement compared to the current situation so I merged it to the feature branch already. Anyway, if you have any comment, please add. I will open a new PR for it.

Note: the only two places where I have not refactored are the RasterSourceTable since it contains some other logics to check is any raster layers is connected to the source and the LocationTable/LocationForm since you are currently working on it so to avoid possible conflicts.

Kr, Hoan

Issue #658, opened 2021-02-11
location form


Derryrover:
Proceed here:
https://github.com/nens/lizard-management-client/issues/770

Derryrover:
in progress:
https://github.com/nens/lizard-management-client/pull/756

Derryrover:
@LirryPinter 

Hi Lirry,

I am looking at the design for the asset selection here:
https://xd.adobe.com/view/64bdf742-fee5-4907-a0f1-e9f27ecac3fc-31ee/screen/a25e4561-7aa3-45ba-aaa0-beb61ec40d3b/

There are some similarities to the map field here. Should they work the same?:
https://nxt3.staging.lizard.net/management/#/alarms/notifications/raster_alarms/new

Would it be possible to work out in some more detail the interactions for this field?
Also backend technical it is possible to have a location with a asset (in object field), but the geometry field is not identical to the geometry of the asset. 
I am afraid we need to cover this situation in the frontend too. Maybe not needed that the user can choose a different geometry from the selected asset, but it is probably needed that if they already differ that the user can somehow see that (and recover from it).

In the meantime I progress as follows:
- In case the geometry of selected asset and location differ, user sees a red text under geometry + a button "remove overwritten geometry".
- User sees on  the map all assets that pass the filter. (since many assets exist, I may need to paginate)

kr Tom

Derryrover:
Problem with backend here:
https://github.com/nens/lizard-nxt/issues/4942

I will investigate if it is because I am unauthorized to use this asset for this organisation.

Derryrover:
@lexvand 
@LirryPinter 
@byrman 

Apart from the json field on location, I also did not add the supplier field.
Supplier seems not exposed in the endpoint

Derryrover:
Proceed here:
https://github.com/nens/lizard-management-client/issues/770

Derryrover:
@hoanphungt 

Hi Hoan,

Do you have time to review?:
https://github.com/nens/lizard-management-client/pull/756

I know it is not finished so we can proceed here:
https://github.com/nens/lizard-management-client/issues/770

kr Tom

hoanphungt:
@Derryrover 
Looks good.

Derryrover:
@hoanphungt 
Hi Hoan,

Good that you further picked up this ticket.
If you still need improvements from me  in the pr please let me know.

kr Tom

PR #785, opened 2021-04-12
Location form - part 2

hoanphungt:


src/data_management/timeseries/locations/LocationForm.tsx
```diff
@@ -389,124 +207,36 @@ const LocationFormModel = (props:Props & PropsFromDispatch & RouteComponentProps
...
-              <FormActionButtons
-                actions={[
-                  {
-                    displayValue: "Delete",
-                    actionFunction: () => {setShowDeleteModal(true)}
-                  },
-                ]}
-              />
-            </div>
-            :null}
```
hoanphungt, line 472: Removed the FormAction for now since this is also not supported in other forms. We can make a ticket to add the FormAction to all the forms if neccessary.

src/data_management/timeseries/locations/LocationForm.tsx
```diff
@@ -80,91 +40,52 @@ const LocationFormModel = (props:Props & PropsFromDispatch & RouteComponentProps
...
-      };
-    const geometryRelatedAsset = 
-      relatedAsset && 
-      relatedAsset.view && 
-      relatedAsset.view[1] !== undefined && 
-      relatedAsset.view[0] !== undefined &&
-      {
-        lat: relatedAsset.view[1],
-        lng: relatedAsset.view[0],
-      };
```
hoanphungt, line 92: I am not covering the geometry of related asset for now and get the geometry directly from the location to simplify this a bit. I will discuss with Lex further about it.

src/alarms/notifications/raster_alarms/RasterAlarmForm.tsx
```diff
@@ -209,6 +209,7 @@ const RasterAlarmForm: React.FC<Props & DispatchProps & RouteComponentProps> = (
           errorMessage={required('Please select a raster', values.raster)}
           triedToSubmit={triedToSubmit}
           isAsync
+          isCached
```
hoanphungt, line 212: Added this optional parameter to the SelectDropdown to specify if cache option is activated for the async select dropdown. The SelectDropdown for asset selection which cannot be cached is the reason I added this parameter.

src/form/MapSelectAssetOrPoint.tsx
```diff
@@ -78,61 +81,55 @@ const MapSelectAssetOrPoint = (props:Props) => {
...
+  //     //   );
+  //     //   if (!inBounds) { 
+  //     //     return; 
+  //     //   }
+  //     // }
+  //     valueChanged({ asset: asset, location: {lat ,lng}});
+  //   } else {
+  //     valueChanged({ asset: null, location: null});
+  //   }
+  // };
```
hoanphungt, line 123: Commented these codes out for now. After this one is merged, I will continue with the ticket to make this component reusable by the RasterAlarmForm and the GeometryField as well.

src/data_management/timeseries/locations/LocationForm.tsx
```diff
@@ -294,93 +173,32 @@ const LocationFormModel = (props:Props & PropsFromDispatch & RouteComponentProps
...
-        {/* <TextArea
-          title={'Extra metadata (JSON) *'}
+        <TextArea
+          title={'Extra metadata (JSON)'}
           name={'extraMetadata'}
-          placeholder={'Enter valid JSON'}
+          placeholder={'Please enter in valid JSON format'}
           value={values.extraMetadata}
           valueChanged={handleInputChange}
           clearInput={clearInput}
```
Derryrover, line 204: The JSON field can for now be commented out since there is a bug with it:
https://github.com/nens/lizard-management-client/issues/658#issuecomment-813976038

src/alarms/notifications/raster_alarms/RasterAlarmForm.tsx
Derryrover: Probably good. I do not yet understand why Asset selection cannot be cached.
Could you explain a bit?

hoanphungt: Yes, so basically, the search endpoint for assets includes a type parameter, this type parameter can be changed using the asset type dropdown. If the async search is cached then even if the asset type is changed, previous search is still cached and will give us outdated list of assets.

src/data_management/timeseries/locations/LocationForm.tsx
hoanphungt: I check your comment in the other ticket. However, there is no mention of why the JSON field is commented out and which bug it is associated with. Could you check again? I can use the JSON field with no problem.

hoanphungt: The supplier field is indeed not supported by the API yet so I removed it completely.

src/form/GeometryField.tsx
```diff
@@ -82,59 +42,66 @@ export const GeometryField: React.FC<MyProps> = (props) => {
...
-            valueChanged={(value)=>{
-              if (isNaN(value)) {
-                setLng(NaN);
-              } else {
-                valueChangedLng(value);
-              }
-              
+            value={value.location ? value.location.lng : NaN}
+            valueChanged={e => {
+              if (isNaN(e)) return;
```
Derryrover, line 59: Not super important, but now when the user empties the field it becomes 0. Previously the field was allowed to be really empty.

hoanphungt: Yes, I will fix this

Derryrover (approved):

src/data_management/timeseries/locations/LocationForm.tsx
hoanphungt: commented out the JSON field

PR #814, opened 2021-05-06
mapviewer tile

Derryrover:


hoanphungt (approved):
react-map-gl looks good! a good start. One note is that I got a lot of CORS errors when zooming in and out of the map with a layer on.

PR #840, opened 2021-05-27
Add invitation modal to user table

hoanphungt:


src/users/InvitationModal.tsx
```diff
@@ -0,0 +1,142 @@
...
+      >
+        <Table
+          tableData={invitations}
+          setTableData={() => null}
+          gridTemplateColumns={'60fr 30fr 10fr'}
+          columnDefinitions={columnDefinitions}
+          dataRetrievalState={dataRetrievalState}
+          triggerReloadWithCurrentPage={() => null}
+          triggerReloadWithBasePage={()=> null}
+          responsive
```
Derryrover, line 122: An alternative solution would be to use here also the TableStateContainer instead of making the user-table manage this state. Why did you choose here to only use the Table component?

src/components/Table.tsx
```diff
@@ -56,12 +67,14 @@ const Table: React.FC<Props> = ({tableData, setTableData, gridTemplateColumns, c
     <div  className={styles.Table}>
       <div style={{
         gridTemplateColumns: gridTemplateColumns,
-        paddingRight: tableIsOverflow ? scrollbarWidth : undefined
+        paddingRight: tableIsOverflow ? scrollbarWidth : undefined,
+        minWidth: responsive ? 'unset' : undefined,
```
Derryrover, line 71: I think I hardcoded the minwidth previously to 1200 in the css.
We now need this to be more flexible.

I see to possible alternatives for the current solution of making table responsive vs unresponsive:
- pass the minwidth as a parameter to table and tableStateContainer
- make the table+tablestateContainer always width=100% but hardcode the component around it (not sure if this is easy enough in the current setup).

src/users/InvitationModal.tsx
hoanphungt: Indeed, I didn't want to use the TableStateContainer at first because the table here doesn't contain any search bar, new item button and most importantly no pagination. But think about it again, I think it's better to just use the TableStateContainer, so I changed it accordingly.

src/components/Table.tsx
hoanphungt: Your solution also sounds good but I will keep the responsive parameter for now.

PR #856, opened 2021-06-09
Flush data from raster source

hoanphungt:


src/components/DeleteModal.tsx
```diff
@@ -65,7 +65,8 @@ function DeleteModal (props: MyProps & DispatchProps & RouteComponentProps) {
       buttonConfirmName={'Delete'}
       onClickButtonConfirm={handleDelete}
       cancelAction={props.handleClose}
-      disableButtons={busyDeleting}
+      disabledCancelAction={busyDeleting}
+      disabledConfirmAction={busyDeleting}
```
hoanphungt, line 69: I need to separate them into 2 parameters as sometimes the confirm button needs to be disabled but not the cancel action.

src/data_management/rasters/RasterSourceTable.tsx
```diff
@@ -2,7 +2,6 @@ import React, { useState, useEffect } from 'react';
 import { useSelector} from 'react-redux';
```
hoanphungt, line 2: I refactor this RasterSourceTable component quite a bit since I want to reuse the DeleteModal and refactor the deleteAction (this work has already been done for all other tables but not this one since this case is a bit more complicated due to its related raster layers). Also, many of commented out code is no longer relevant so I removed them too.

In the old code, there is still a case of a failed deletion of raster source with a **412** status response which is not handled by the DeleteMoal. I could not reproduce this specific case so do you know how important it is and if we should still handle this case? If yes, I can continue improve this part.

Derryrover (approved):

PR #887, opened 2021-07-14
Landing page styling

Derryrover:


src/home/AppTileConfig.ts
```diff
@@ -65,12 +68,16 @@ export const getNavigationLinkTileFromUrlAndAllNavigationLinkTiles = (urlPostFix
   }
 }
 
+export const getCurrentUrlPostfix = () => {
+  return window.location.href.split("/#")[1];
+}
```
hoanphungt, line 73: This function seems hacky and do we still need the # for the new url anymore? I thought the # was used previously because we have a django admin page that also lives in the /management domain. Now this page is no longer existed, we can later configure to not have the #

hoanphungt (approved):

src/home/AppTileConfig.ts
Derryrover: Good point!
I made this ticket:
https://github.com/nens/lizard-management-client/issues/894

PR #896, opened 2021-07-21
Add react query

hoanphungt:


hoanphungt:
@Derryrover 
I made a custom hook useRecursiveFetch based on useQuery to replace the previous usePaginatedFetch. I think it's a good place to start using react-query. Could you review?

src/api/hooks.ts
```diff
@@ -0,0 +1,110 @@
...
+    queries: {
+      staleTime: 5 * 60 * 1000,
+      retry: false as const,
+      refetchOnMount: false as const,
+      // @ts-ignore
+      refetchOnWindowsFocus: false as const,
+      refetchOnReconnect: true as const,
+      refetchIntervalInBackground: true as const,
+      refetchInterval: false as const,
+    }
```
Derryrover, line 19: I like the overall solution, but wonder what will happen if we need another "staleTime" at another query.
Should we maybe not have the "QueryClientProvider" at rootlevel?
Or should we just give all queries a staleTime of 0 ?
Or do we not use the react-query for these not-tobe-cached api calls and use normal fetch instead?
What are your thoughts on this ?

hoanphungt: It is possible to adjust staleTime to 0 for another query by specifying it as below:
`const query = useQuery(key, fetchFunction, { staleTime: 0 })`

In my opinion, we mostly want fresh data for the app so I think the default staleTime should be 0 (also the default value by react-query so I can just remove this option). For a specific request that we want the results to be cached then we can set the staleTime value like above (I can't think of any example right now)

PR #908, opened 2021-07-26
confirm leave show upload queue on stay

Derryrover:


remcogerlich (approved):
Read through the changes and see nothing I object to, so fine with me

PR #931, opened 2021-08-19
Tom landing page scenario usage

Derryrover:


src/actions.js
```diff
@@ -96,21 +99,40 @@ export function fetchOrganisations() {
...
+            type: SET_CONTRACTS,
+            contracts: data,
+          });
+          if (data.next) {
+            const relativeUrl = data.next.split('lizard.net')[1];
+            requestContracts(relativeUrl);
+          }
+      });
+    };
+    requestContracts(`/api/v4/contracts`);
```
hoanphungt, line 123: I made a `recursiveFetchFunction` in the hooks.ts file. I think it can be used here too. An example of where it is used is the fetchOrganisations function in action.ts.

src/components/NoOrganisationFoundComponent.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import agreementIcon from "../images/agreement.svg";
+
+
+const NoOrganisationFoundComponent = () => {
```
hoanphungt, line 5: Should the name be like: NoContractFoundComponent instead?

src/components/UsagePieChart.tsx
```diff
@@ -0,0 +1,86 @@
...
+            ></circle>
+            <circle
+              cx="0" 
+              cy="0" 
+              r="15.91549430918952" 
+              fill="none"
+              stroke="#139696" 
+              stroke-width="4"
+              stroke-dasharray={`${percentageUsed} ${percentageNotUsed}`} 
+              stroke-dashoffset="25"
```
hoanphungt, line 54: I got multiple warnings (3) here in my console: `stroke-width` should be `strokeWidth`, `stroke-dasharray` should be `strokeDasharray` and `stroke-dashoffset` should be `strokeDashoffset`

src/components/UsagePieChart.tsx
```diff
@@ -0,0 +1,86 @@
...
+          >
+            <circle
+              cx="0" 
+              cy="0" 
+              // use this radius see tutorial here:
+              // https://heyoka.medium.com/scratch-made-svg-donut-pie-charts-in-html5-2c587e935d72
+              r="15.91549430918952" 
+              fill="none"
+              stroke="#E6E6E6" 
+              stroke-width="4"
```
hoanphungt, line 44: strokeWidth instead of stroke-width to avoid console warning

src/components/UsagePieChart.tsx
```diff
@@ -0,0 +1,86 @@
...
+            position: "relative",
+            alignItems: "flex-start",
+            display: "inline-flex",
+            flexDirection: "column",
+          }}
+        >
+          <svg
+            height={138}
+            width={138}
+            viewBox={"-18 -18 36 36"}
```
hoanphungt, line 34: I think this svg for the donut chart can preferably be made into a reusable component (in ts file like in the 3di so it can also receive props). But since the donut chart is being used only here so it's up to you whether you think it's useful to make it reusable.

src/contracts/ContractForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+      <form
+        className={formStyles.Form}
+        onSubmit={handleSubmit}
+        onReset={handleReset}
+      >
+        <span className={`${formStyles.FormFieldTitle} ${formStyles.FirstFormFieldTitle}`}>
+          1: General
+        </span>
+        <TextInput
+          title={'Start dat contract'}
```
hoanphungt, line 59: spelling: `date` instead of `dat`

src/contracts/ContractForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+          name={'start_date_contract'}
+          value={(new Date(contractObjApi.start).toLocaleDateString())}
+          valueChanged={handleInputChange}
+          clearInput={clearInput}
+          validated={!minLength(2, values.name)}
+          errorMessage={minLength(2, values.name)}
+          triedToSubmit={triedToSubmit}
```
hoanphungt, line 66: I think these props can be removed since this is only a read-only field so no validation is needed (only use `validated`)

src/contracts/ContractForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+          onFocus={handleFocus}
+          onBlur={handleBlur}
+          readOnly={true}
```
hoanphungt, line 69: can be just `readOnly`

src/contracts/ContractForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+        </span>
+        <TextInput
+          title={'Assets capacity'}
+          name={'asset_capacity'}
+          value={contractObjApi.asset_capacity}
+          valueChanged={handleInputChange}
+          clearInput={clearInput}
+          validated={!minLength(2, values.name)}
+          errorMessage={minLength(2, values.name)}
+          triedToSubmit={triedToSubmit}
```
hoanphungt, line 120: same as above (can be removed and only use validated)

src/contracts/ContractForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+        />
+        <TextInput
+          title={'Labels capacity'}
+          name={'label_capacity'}
+          value={contractObjApi.label_capacity}
+          valueChanged={handleInputChange}
+          clearInput={clearInput}
+          validated={!minLength(2, values.name)}
+          errorMessage={minLength(2, values.name)}
+          triedToSubmit={triedToSubmit}
```
hoanphungt, line 133: same as above

src/contracts/ContractForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+        />
+        <TextInput
+          title={'Events capacity'}
+          name={'event_capacity'}
+          value={contractObjApi.event_capacity}
+          valueChanged={handleInputChange}
+          clearInput={clearInput}
+          validated={!minLength(2, values.name)}
+          errorMessage={minLength(2, values.name)}
+          triedToSubmit={triedToSubmit}
```
hoanphungt, line 146: same as above

src/contracts/ContractForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+        />
+        <TextInput
+          title={'Alarm message capacity'}
+          name={'alarm_message_capacity'}
+          value={contractObjApi.alarm_message_capacity}
+          valueChanged={handleInputChange}
+          clearInput={clearInput}
+          validated={!minLength(2, values.name)}
+          errorMessage={minLength(2, values.name)}
+          triedToSubmit={triedToSubmit}
```
hoanphungt, line 159: same as above

src/contracts/ContractForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+        <CheckBox
+          title={'Geoblocks functionality'}
+          name={'geoblocks_functionality'}
+          value={contractObjApi.geoblocks_acces}
+          valueChanged={bool => handleValueChange('scopeWildcardReadWrite', bool)}
+          onFocus={handleFocus}
+          onBlur={handleBlur}
+          readOnly={true}
+          validated={true}
+          errorMessage={false}
```
hoanphungt, line 173: errorMessage can be removed since validated is always true

src/contracts/ContractForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+        />
+        <TextInput
+          title={'Geoblocks calculation units'}
+          name={'geoblocks_calculation_units'}
+          value={contractObjApi.geoblocks_calculation_units_capacity}
+          valueChanged={handleInputChange}
+          clearInput={clearInput}
+          validated={!minLength(2, values.name)}
+          errorMessage={minLength(2, values.name)}
+          triedToSubmit={triedToSubmit}
```
hoanphungt, line 183: same as above

src/contracts/ContractForm.tsx
```diff
@@ -0,0 +1,212 @@
...
+        />
+        <TextArea
+          title={'Links'}
+          name={'links'}
+          placeholder={'This is a layer based on raster_source'}
+          value={contractObjApi.links}
+          valueChanged={handleInputChange}
+          onFocus={handleFocus}
+          onBlur={handleBlur}
+          clearInput={clearInput}
```
hoanphungt, line 196: clearInput can be removed for readOnly field

src/reducers.js
```diff
@@ -108,6 +115,24 @@ switch (action.type) {
   } 
 }
 
+function contracts (
+  state = {
+    contracts: [],
+    isFetching: false,
+    timesFetched: 0,
```
hoanphungt, line 122: why do you need to keep this timesFetched?

hoanphungt (changes requested):
Some comments in the PR

src/actions.js
Derryrover: done. Good point actually ;)

src/components/NoOrganisationFoundComponent.tsx
Derryrover: Done, nice catch !

src/components/UsagePieChart.tsx
Derryrover: done

src/components/UsagePieChart.tsx
Derryrover: done

src/contracts/ContractForm.tsx
Derryrover: done

src/contracts/ContractForm.tsx
Derryrover: done

src/components/UsagePieChart.tsx
Derryrover: Good idea to make it more reusable.
I made this ticket:
https://github.com/nens/lizard-management-client/issues/936

PR #951, opened 2021-09-08
[Hoan] visualize geoblock using rete.js

hoanphungt:


hoanphungt:
Decide to use react-flow-renderer instead of rete.js

PR #973, opened 2021-10-04
[DRAFT] Make inbound edges as part of the elements

hoanphungt:


hoanphungt:
This is an experimental pull request.

Conclusion: Make inbound edges as part of the elements inside the createGraphLayout function is possible and make the main code for geoblock cleaner.

PR #993, opened 2021-11-09
Use uuid library to generate uuid for id of a new block

hoanphungt:


hoanphungt:
@Derryrover 
I used uuid library to generate uuid for a new block.

PR #1001, opened 2021-11-16
Fix staled selected organisation from local storage

hoanphungt:


src/actions.js
```diff
@@ -97,7 +97,8 @@ export function fetchOrganisations() {
       const selectedOrganisation = availableOrganisations[0];
       dispatch(selectOrganisation(selectedOrganisation));
     } else {
-      dispatch(selectOrganisation(selectedOrganisationLocalStorage));
+      const selectedOrganisation = availableOrganisations.find(org => org.uuid === selectedOrganisationLocalStorage.uuid);
```
Derryrover, line 100: Hi Hoan,
I agree that this must fix the reported bug:
It retrieves the latest available organisation-roles and uses this in the app.
(as opposed to that it uses some historically stored data)

This is a bit hard to test for me now, so I did not test it.

kr Tom

Derryrover (approved):

src/actions.js
hoanphungt: No worries, I am gonna test it again when the fix is on staging. I tested it on dev by changing my own rights in the admin page.

PR #1041, opened 2022-02-24
Test vector tiles

hoanphungt:


hoanphungt:
@Derryrover 
I changed some of the MapViewer comonent to display the measuring stations as vector tiles in the map. It's currently on staging. You can go to the Map Viewer to have a look at.

PR #1049, opened 2022-03-17
Test maplibre with react-map-gl

hoanphungt:


hoanphungt:
Conclusion: react-map-gl works well with mapLibre

PR #1050, opened 2022-03-22
Test maplibre with reac-map-gl

hoanphungt:


hoanphungt:
Conclusion: react-map-gl v7 works well with mapLibre and typescript

PR #1051, opened 2022-04-05
Animate temporal raster using react-leaflet

hoanphungt:


hoanphungt:
@Derryrover @remcogerlich 
Hi Tom and Remco,

I made a animation with react-leaflet for the regen layer (the same result can probably be achieved using react-map-gl).

The idea is I pre-loaded all the tiles during the whole time frame and waited for it to finish before playing the animation. All the layers will be placed on the map but at each time frame, only one layer with the correct timestamp will be shown up (by setting `opacity` from 0 to 1).

For each timestep, I also included a `loaded` property to know when all the tiles of that layer have been loaded using the `onload` event. Whenever zoom in/zoom out, it will trigger the `ontileloadstart` event which will stop the animation and wait for tiles to be reloaded before you can play the animation again.

So it's quite a lot of a hassle to play an animation using this approach I think but it is still a step forward :). Please try to play the animation and let me know what you guys think.

Derryrover:
@hoanphungt 

Hi Hoan,

I think it looks good!
I think this is overall a approach we could take.
We could think about:
- Will we use mapboxGLJS instead of leaflet?
- Do we always want to preload a certain amount of times or could we also just load once needed? Right now we need to have a "static" end time of the animation.
- Could we use TMS instead of WMS? Will this help in performance or not?
- Which completely other strategies could we use?

kr Tom

hoanphungt:
I forgot to mention but you can test it on dev by going to the MapViewer

hoanphungt:
@Derryrover To answer your points:

- It is possible to use mapbox-gl instead of leaflet for the animation. I already made a PR using mapbox-gl, I will update that PR to see if it gives us the same effect compared to leaflet.
- I think the smoothest approach is to pre-load tiles for all timesteps. Currently, the lizard-viewer preloads a few timesteps ahead, not all.
- TMS could be a good alternative to WMS in term of browser caching the tiles. I read online and it is said that TMS gives a predefined set of bboxes/scales so they can be cached quite effectively compared to WMS. But it's a step that requires the backend to provide us with TMS layers. I already talked with @arjanverkerk about it and if I remember correctly, we already had some existing codes for the TMS. So, in conclusion, yes, TMS could be a good option to try.
- Other strategies? Needs to investigate more.

src/components/Mapviewer.tsx
```diff
@@ -1,161 +1,129 @@
...
+      <div>
+        <button
+          onClick={() => startAnimation()}
+          disabled={!tilesReady || playAnimation}
+        >
+          {tilesReady ? 'Play' : 'Loading ...'}
+        </button>
+        {' '}
+        <button
+          onClick={() => stopAnimation()}
```
remcogerlich, line 120: As stopAnimation is already a function with 0 arguments, this is the same as `onClick={stopAnimation}`, no need to create yet another function. (But onClick will send the event as first argument, so watch out with functions where that would be a problem). Same for startAnimation.

src/components/Mapviewer.tsx
```diff
@@ -1,161 +1,129 @@
...
+        ))}
+      </Map>
+      <div>
+        <button
+          onClick={() => startAnimation()}
+          disabled={!tilesReady || playAnimation}
+        >
+          {tilesReady ? 'Play' : 'Loading ...'}
+        </button>
+        {' '}
```
remcogerlich, line 118: What is this space for?

src/components/Mapviewer.tsx
```diff
@@ -1,161 +1,129 @@
...
+            // the onTileLoadStart happens when a tile is requested and starts loading
+            // in this case, stop the animation and set the "loaded" parameter back to false
+            // to indicate that the tiles are being reloaded
+            // However, it normally takes very long to reload all the tiles (way longer than the first time)
+            // During this time, user is not supposed to zoom in/zoom out, else the reloading process will happen again
+            ontileloadstart={() => {
+              stopAnimation()
+              setTimestamps(timestamps.map(ts => ({
+                ...ts,
+                loaded: false
```
remcogerlich, line 105: This seems like a problem, it sets _all_ the loaded things to false, for all timestamps? And each timestamp has its own ontileloadstart?

Should probably pass a function to setTimestamp, that takes the current value and then only sets the right one, to prevent race conditions later.

remcogerlich: Also I'm not sure it's acceptable to stop animating whenever the user pans or zooms.

src/components/Mapviewer.tsx
```diff
@@ -1,161 +1,129 @@
...
+            key={timestamp.time}
+            url={'/wms/'}
+            layers={'radar:5min'}
+            styles={'radar-5min'}
+            time={timestamp.time}
+            format={'image/png'}
+            uppercase={true}
+            bounds={getBounds(raster)}
+            tileSize={1024}
+            opacity={timestamp.time === timestamps[step].time ? 1 : 0}
```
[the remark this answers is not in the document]
hoanphungt: Yes, I saw it too in the WaterDepthImageLayer component and you set the visibility in the Pane. In my case, I checked and there is no difference between using visibility and opacity (both are smooth), but that's because I preloaded all the tiles.

src/components/Mapviewer.tsx
hoanphungt: noted

src/components/Mapviewer.tsx
hoanphungt: ah just to separate two buttons next to each other

src/components/Mapviewer.tsx
hoanphungt: You are right. It could introduce a race conditions. I fixed it. 

For zooming and panning issue, it could pause to reload instead of stop and then resume.

PR #1069, opened 2022-06-08
Add scenario result form

hoanphungt:


hoanphungt:
Close this PR due to another solution used in #1070

